PR #10, opened 2024-05-31
Authentication spec for ORCA application

rolandgroen:
Added two new files detailing the user authentication process for the ORCA application from the Viewer application. The files include a UML sequence diagram in puml format and a markdown file that explains the process in detail. The process utilizes OIDC flow with the id_token_hint option, also outlines potential future authentication requirements.


reinkrul (approved):
Looks like an approach that should work. OK for initial version, we'll need more refined documentation soon.

reinkrul (changes requested):
Things have changed, we're not going to require the browser to be present for authentication on the other side.

Meaning, service access token flow spec, with the Viewer's IdP id_token in some form (attributes, as-is or Verifiable Credential).

PR #27, opened 2024-06-27
Auto enrich task

JorisHeadease:
INT-129: After a task has been accepted by the filler, the placer will now enrich the task with Task.includes and Task.contains and also the Task.for


orchestrator/careplancontributor/service.go
```diff
@@ -104,24 +106,63 @@ func (s Service) confirm(localFHIR fhirclient.Client, serviceRequestRef string,
...
+
+	// Create channels for polling the task, 1 channel for the result and 1 for an error
+	resultChan := make(chan *fhir.Task)
+	errorChan := make(chan error)
+
+	// Start polling in a new goroutine so that the code continues to the select below
+	go s.pollTaskStatus(*task.Id, resultChan, errorChan)
+
+	// Block until either the task is accepted or an error occurs (error after 10 attempts)
+	select {
```
reinkrul, line 122: this makes the call effectively blocking again, so no need to do `pollTaskStatus` in a goroutine

orchestrator/careplancontributor/service.go
```diff
@@ -161,8 +202,8 @@ func (s Service) readServiceRequest(localFHIR fhirclient.Client, serviceRequestR
 func (s Service) createCarePlan(patient fhir.Patient) (*fhir.CarePlan, error) {
 	carePlan := fhir.CarePlan{
 		Subject: fhir.Reference{
-			Type:       to.Ptr("Patient"),
-			Identifier: &patient.Identifier[0], // TODO: is this the right way/one?
+			Type:      to.Ptr("Patient"),
+			Reference: to.Ptr(*patient.Id),
```
reinkrul, line 206: `Id` is already a pointer, no need to dereference and make a pointer again
```suggestion
			Reference: patient.Id,
```

orchestrator/careplancontributor/service.go
```diff
@@ -104,24 +106,63 @@ func (s Service) confirm(localFHIR fhirclient.Client, serviceRequestRef string,
...
+			err := fmt.Errorf("maximum number of attempts reached for Task/%s", taskID)
+			log.Error().Msgf("polling Task/%s failed - error: [%s]", taskID, err)
+			errorChan <- err
+			return
+		}
+
+		attempts++
+		var task fhir.Task
+
+		if err := s.CarePlanService.Read("Task/"+taskID, &task); err != nil {
```
reinkrul, line 153: This doesn't take into account the time a `Read` takes. So maybe rewrite the `for` loop with a `time.Sleep(..)` instead, and check whether the expected total time taken (now 20s) has passed? You could also do this with the ticker, but then with a `for { select (ticker.C, ctx.Done) }` (pseudo code) instead, that uses a context passed to the`pollTaskStatus` function (which can then have a time-out).

(https://golang.cafe/blog/golang-context-with-timeout-example.html)

Also; `Read` invokes an HTTP client, which has a default time-out of 30s. So this could the first iteration to time-out, right at the beginning. So ideally we specify a time-out in the `Read()` function (or actually a context which you can cancel/give a time-out). I'll add that to the library.

orchestrator/careplancontributor/service.go
```diff
@@ -172,43 +213,89 @@ func (s Service) createCarePlan(patient fhir.Patient) (*fhir.CarePlan, error) {
...
+			},
+		},
+	}
+
+	taskMap["contained"] = []interface{}{
+		*serviceRequest,
+		*patient,
+	}
+
+	taskJSON, _ = json.Marshal(taskMap)
```
reinkrul, line 294: the FHIR library already does (un)marshalling, so no need to marshal here. Just return the `Task` map.

orchestrator/careplancontributor/service.go
```diff
@@ -172,43 +213,89 @@ func (s Service) createCarePlan(patient fhir.Patient) (*fhir.CarePlan, error) {
...
+// When an application accepts the Task, we enrich the Task with more detailed information like the actual Patient and the ServiceRequest
+func (s Service) handleAcceptedTask(task *fhir.Task, serviceRequest *fhir.ServiceRequest, patient *fhir.Patient) (*fhir.Task, error) {
+
+	taskJSON, err := s.enrichTask(task, serviceRequest, patient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to enrich task: %w", err)
+	}
+
+	var enrichedTask fhir.Task
+	if err := s.CarePlanService.Update("Task/"+*task.Id, json.RawMessage(*taskJSON), &enrichedTask); err != nil {
```
reinkrul, line 246: if `enrichTask` returns the `Task` (as map), you can simply pass it here and let the FHIR client do the marshalling).

reinkrul (changes requested):
Maybe we can also start writing unit tests here and there for stuff that's somewhat final?

orchestrator/careplancontributor/service.go
JorisHeadease: Rewrote the logic so that the complete process times out after 20 seconds. I've added a todo to add a timeout to the Read when the `go-fhir-client` supports it.

orchestrator/careplancontributor/service.go
```diff
@@ -111,57 +111,42 @@ func (s Service) confirm(localFHIR fhirclient.Client, serviceRequestRef string,
...
-			errorChan <- err
-			return
+func (s Service) pollTaskStatus(taskID string) error {
+	pollInterval := 2 * time.Second
+	maxPollingDuration := 20 * time.Second
+	startTime := time.Now()
+
+	for {
+		if time.Since(startTime) >= maxPollingDuration {
+			return fmt.Errorf("maximum polling duration of %d seconds reached for Task/%s", maxPollingDuration, taskID)
```
reinkrul, line 132: I think it will print it in nanoseconds now, better:
```suggestion
			return fmt.Errorf("maximum polling duration of %s reached for Task/%s", maxPollingDuration, taskID)
```

(will print `20s`)

orchestrator/careplancontributor/service.go
```diff
@@ -111,57 +111,42 @@ func (s Service) confirm(localFHIR fhirclient.Client, serviceRequestRef string,
...
 		}
 
-		attempts++
 		var task fhir.Task
 
+		//TODO: Add timeout to the read when the lib supports it
 		if err := s.CarePlanService.Read("Task/"+taskID, &task); err != nil {
-			errorChan <- err
 			log.Error().Msgf("polling Task/%s failed - error: [%s]", taskID, err)
```
reinkrul, line 139:
```suggestion
```

orchestrator/careplancontributor/service.go
```diff
@@ -111,57 +111,42 @@ func (s Service) confirm(localFHIR fhirclient.Client, serviceRequestRef string,
...
-		log.Info().Msgf("Task/%s accepted: %+v\n", *result.Id, result)
-		enrichedTask, err := s.handleAcceptedTask(task, serviceRequest, &patient)
-		return enrichedTask, err
-	case err := <-errorChan:
-		fmt.Printf("error while polling task %s: %v\n", *task.Id, err)
-		return nil, err
+	err = s.pollTaskStatus(*task.Id)
+
+	if err != nil {
+		return nil, fmt.Errorf("error while polling task %s: %v\n", *task.Id, err)
```
reinkrul, line 118: Wrap the error with `%w`:
```suggestion
		return nil, fmt.Errorf("error while polling task %s: %w", *task.Id, err)
```

reinkrul (approved):
Better. Small things left.

PR #33, opened 2024-07-17
INT-130 - Update CareTeam on Task update

reinkrul:
This PR adds functionality to CPS to update the CareTeam when a FHIR Task is updated. It updates membership according to https://build.fhir.org/ig/santeonnl/shared-care-planning/branches/main/overview.html#updating-careplan-and-careteam

TODO:
- [x] Update CareTeam on Task creation

Things to work out:
- [ ] Do we want to remove `CareTeam.participants` that we don't have a Task for? In case we want to support CareTeams with participants that don't have an actual Task? (@bramwesselo ?) (out of scope for now)


orchestrator/lib/coolfhir/util.go
```diff
@@ -26,8 +42,28 @@ func FirstIdentifier(identifiers []fhir.Identifier, predicate func(fhir.Identifi
...
-func IsNamingSystem(system string) func(fhir.Identifier) bool {
+func FilterNamingSystem(system string) func(fhir.Identifier) bool {
 	return func(ident fhir.Identifier) bool {
 		return ident.System != nil && *ident.System == system
 	}
 }
+
+func ValidateLogicalReference(reference *fhir.Reference, expectedType string, expectedSystem string) error {
+	if reference == nil {
+		return errors.New("not a reference")
```
dwelman, line 53: "reference cannot be null" might be more descriptive?

dwelman (approved):
Looks good, minor nitpick but can be ignored if you feel what we have is good already

orchestrator/careplanservice/integration_test.go
```diff
@@ -42,29 +44,58 @@ func Test_Integration_Service_handleCreateTask(t *testing.T) {
...
+				})
+				t.Run("Check that CarePlan.activities contains the Task", func(t *testing.T) {
+					err = carePlanContributor.Read("CarePlan/"+*carePlan.Id, &carePlan)
+					require.NoError(t, err)
+					require.Len(t, carePlan.Activity, 1)
+					require.Equal(t, "Task", *carePlan.Activity[0].Reference.Type)
+					require.Equal(t, "Task/"+*task.Id, *carePlan.Activity[0].Reference.Reference)
+				})
+
+				t.Run("Accept Task", func(t *testing.T) {
```
dwelman, line 75: Should this not be in method `Test_Integration_Service_handleUpdateTask`? 

Otherwise, if it is more convenient to test the update after the create, maybe we can rename the test method `Test_Integration_Service_handleCreateTask` something more generic/applicable like `Test_Integration_Service_handleCRUDTask` or similar?

reinkrul: This had become a bit messy, I'd like to have 1 integration test for Task lifecycle (for now), so I'll just skip on nesting tests, and just make it 1 big test.

reinkrul:
I'm going to rework this to have it watch the actual state transition (participant is only added to CareTeam when transition to "accepted" is made), instead of building IST and SOLL (current state and intended state) and  making changes based on that. The latter method is much more complicated and doesn't support the `ready -> completed` (no CareTeam membership) transition.

orchestrator/careplanservice/integration_test.go
dwelman: Sounds good

PR #43, opened 2024-08-23
INT-238: Implement state transition validation for Task in CPS

dwelman:
Adds all necessary transition logic and testing for transitions between all states.

Future work to be done: Implement the update API and handler


orchestrator/careplanservice/handle_updatetask.go
```diff
@@ -0,0 +1,87 @@
...
+				break
+			}
+		}
+	}
+	println("isOwner: ", isOwner)
+	// (same for requester)
+	return nil
+}
+
+func isValidTransition(from fhir.TaskStatus, to fhir.TaskStatus, isOwner bool, isRequester bool) bool {
```
reinkrul, line 31: check it against the table, looks good

orchestrator/careplanservice/service.go
```diff
@@ -95,6 +95,15 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 			return
 		}
 	}))
+	mux.HandleFunc("PUT /cps/Task/:id", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
```
reinkrul, line 98: actually, it should be `{id}`

dwelman:
Messed up the rebase and the commit history now has a lot of changes I did not make, going to close this PR and create a new branch with the new changes

PR #45, opened 2024-08-29
INT-238: Implement state transition validation for Task update in CPS

dwelman:
When updating a task in CarePlanService there are only certain status transitions that are valid, otherwise return error.

Adds unit tests and updates integration test

To follow: Fetch owner and requester of task to validate that they are allowed to make the status change

### DIFF
--- a/orchestrator/careplanservice/handle_createtask.go
+++ b/orchestrator/careplanservice/handle_createtask.go
@@ -23,8 +23,6 @@ func (s *Service) handleCreateTask(httpResponse http.ResponseWriter, httpRequest
 		return fmt.Errorf("invalid Task: %w", err)
 	}
 	switch task["status"] {
-	case fhir.TaskStatusDraft.String():
-		task["status"] = fhir.TaskStatusRequested.String()
 	case fhir.TaskStatusRequested.String():
 	case fhir.TaskStatusReady.String():
 	default:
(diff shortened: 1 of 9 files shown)


orchestrator/careplanservice/handle_createtask.go
```diff
@@ -22,6 +22,14 @@ func (s *Service) handleCreateTask(httpResponse http.ResponseWriter, httpRequest
 	if err := s.readRequest(httpRequest, &task); err != nil {
 		return fmt.Errorf("invalid Task: %w", err)
 	}
+	switch task["status"] {
+	case fhir.TaskStatusDraft.String():
+		task["status"] = fhir.TaskStatusRequested.String()
```
dwelman, line 27: @bramwesselo is this a good default or should it rather be `ready`?

reinkrul (changes requested):
Small remarks

reinkrul (approved):

orchestrator/careplanservice/handle_createtask.go
bramwesselo: Status = draft should never occur (should raise an error) because it's not an allowed state in the SCPTask profile: https://santeonnl.github.io/shared-care-planning/ValueSet-scp-task-status.html

dwelman: Fixed

PR #47, opened 2024-09-04
INT-237: Removed `NEXT_PUBLIC` env vars

JorisHeadease:
`NEXT_PUBLIC` en vars are baked into the application build artifact during the Docker image creation. Therefore, we cannot change the values when deploying to different environments.

Simplified the setup, proxied fhir requests so there is no need for a "regular" and "docker" config for local development. `NEXT_PUBLIC_BASE_PATH` is still public as this is fixed over all deployments. It can be easily changed by keeping it in the config.


hospital_simulator/.env.production
```diff
@@ -1,3 +1,2 @@
-NEXT_PUBLIC_FHIR_BASE_URL=http://localhost:9090/fhir
-NEXT_PUBLIC_FHIR_BASE_URL_DOCKER=http://fhirstore:8080/fhir
-NEXT_PUBLIC_ORCA_BASE_URL=/orca
\ No newline at end of file
+FHIR_BASE_URL=http://fhirstore:8080/fhir
```
reinkrul, line 1: This is going to be depending on the deployment environment.

reinkrul (changes requested):
see comment

hospital_simulator/.env.production
JorisHeadease: Yeah, this will just be the default for the `Dockerfile` / dev deployment. When Ops overwrites the env var it should overwrite this value.

PR #48, opened 2024-09-09
Return OperationOutcome instead of plain-text Error

tader:
[INT-255: Return OperationOutcome on error](https://linear.app/zorg-bij-jou/issue/INT-255/orca-or-return-operationoutcome-on-error-and-parse-them-client-side)

### DIFF
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -90,7 +90,6 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("POST /cps/Task", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleCreateTask(writer, request)
 		if err != nil {
-			log.Info().Msgf("CarePlanService/CreateTask failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/CreateTask", writer)
 			return
 		}
@@ -98,7 +97,6 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("PUT /cps/Task/{id}", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleUpdateTask(writer, request)
 		if err != nil {
-			log.Info().Msgf("CarePlanService/UpdateTask failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/UpdateTask", writer)
 			return
 		}
@@ -106,7 +104,6 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("POST /cps/CarePlan", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleCreateCarePlan(writer, request)
 		if err != nil {
-			log.Info().Msgf("CarePlanService/CreateCarePlan failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/CreateCarePlan", writer)
 			return
 		}
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -90,6 +90,7 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("POST /cps/Task", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleCreateTask(writer, request)
 		if err != nil {
+			log.Info().Msgf("CarePlanService/CreateTask failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/CreateTask", writer)
 			return
 		}
@@ -97,6 +98,7 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("PUT /cps/Task/{id}", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleUpdateTask(writer, request)
 		if err != nil {
+			log.Info().Msgf("CarePlanService/UpdateTask failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/UpdateTask", writer)
 			return
 		}
@@ -104,6 +106,7 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("POST /cps/CarePlan", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleCreateCarePlan(writer, request)
 		if err != nil {
+			log.Info().Msgf("CarePlanService/CreateCarePlan failed: %v", err)
 			s.writeOperationOutcomeFromError(err, "CarePlanService/CreateCarePlan", writer)
 			return
 		}
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -3,6 +3,10 @@ package careplanservice
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	fhirclient "github.com/SanteonNL/go-fhir-client"
 	"github.com/SanteonNL/nuts-policy-enforcement-point/middleware"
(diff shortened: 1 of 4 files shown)


dwelman (approved):

orchestrator/careplanservice/service_test.go
```diff
@@ -1,17 +1,24 @@
...
...atus</b>:Requested</p><p><b>intent</b>:order</p><p><b>code</b>:<spantitle=\\\"Codes:{http://hl7.org/fhir/CodeSystem/task-codefullfill}\\\">fullfill</span></p><p><b>focus</b>:<ahref=\\\"Bundle-cps-bundle-01.html#urn-uuid-456\\\">Bundle:type=transaction</a></p><p><b>for</b>:Identifier:<code>http://fhir.nl/fhir/NamingSystem/bsn</code>/111222333</p><p><b>requester</b>:Identifier:<code>http://fhir.nl/fhir/NamingSystem/uzi</code>/UZI-1</p><p><b>owner</b>:Identifier:<code>http://fhir.nl/fhir/NamingSystem/ura</code>/URA-2</p><p><b>reasonReference</b>:<ahref=\\\"Bundle-cps-bundle-01.html#urn-uuid-789\\\">Bundle:type=transaction</a></p></div>\"},\"status\":\"requested\",\"intent\":\"order\",\"code\":{\"coding\":[{\"system\":\"http://hl7.org/fhir/CodeSystem/task-code\",\"code\":\"fullfill\"}]},\"focus\":{\"reference\":\"urn:uuid:456\"},\"for\":{\"identifier\":{\"system\":\"http://fhir.nl/fhir/NamingSystem/bsn\",\"value\":\"111222333\"}},\"requester\":{\"identifier\":{\"system\":\"http://fhir.nl/fhir/NamingSystem/uzi\",\"value\":\"UZI-1\"}},\"owner\":{\"identifier\":{\"system\":\"http://fhir.nl/fhir/NamingSystem/ura\",\"value\":\"URA-2\"}},\"reasonReference\":{\"reference\":\"urn:uuid:789\"}}"
```
reinkrul, line 20: could use backticks instead of double quotes to avoid having to escape all the double qoutes to avoid readability. Or put it in a file?

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/samply/golang-fhir-models/fhir-models/fhir"
+)
+
+// Writes an OperationOutcome based on the given error as HTTP response.
+func (s *Service) writeOperationOutcomeFromError(err error, desc string, httpResponse http.ResponseWriter) {
+	diagnostics := fmt.Sprintf("%s failed: %s", desc, err.Error())
+
+	issue := fhir.OperationOutcomeIssue{
+		Id:                nil,
```
reinkrul, line 18: I'd say leave out all the `nil` fields to improve readability

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+		Severity:          fhir.IssueSeverityError,
+		Code:              fhir.IssueTypeProcessing,
+		Details:           nil,
+		Diagnostics:       to.Ptr(diagnostics),
+		Location:          nil,
+		Expression:        nil,
+	}
+
+	outcome := fhir.OperationOutcome{
+		Id:                nil,
```
reinkrul, line 30: same here

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+		ModifierExtension: nil,
+		Issue:             []fhir.OperationOutcomeIssue{issue},
+	}
+
+	httpResponse.Header().Add("Content-Type", "application/fhir+json")
+	httpResponse.WriteHeader(http.StatusBadRequest)
+
+	data, err := json.Marshal(outcome)
+	if err != nil {
+		log.Error().Msgf("Failed to marshal OperationOutcome: %s", diagnostics)
```
reinkrul, line 45: Add `WithError(err)` (or something), otherwise you'll lose the error. You could also choose to ignore the error, since `json.Marshal()` only returns an error if you try to marshal a pointer to a function (which is very, very implausable).

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+		Meta:              nil,
+		ImplicitRules:     nil,
+		Language:          nil,
+		Text:              nil,
+		Extension:         nil,
+		ModifierExtension: nil,
+		Issue:             []fhir.OperationOutcomeIssue{issue},
+	}
+
+	httpResponse.Header().Add("Content-Type", "application/fhir+json")
```
reinkrul, line 40: we could use a constant, maybe in the `coolfhir` package?

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+
+	data, err := json.Marshal(outcome)
+	if err != nil {
+		log.Error().Msgf("Failed to marshal OperationOutcome: %s", diagnostics)
+		return
+	}
+
+	_, err = httpResponse.Write(data)
+	if err != nil {
+		log.Error().Msgf("Failed to return OperationOutcome: %s", diagnostics)
```
reinkrul, line 51: `WithError`

orchestrator/careplanservice/service_test.go
```diff
@@ -46,6 +53,100 @@ func TestService_Proxy(t *testing.T) {
...
+
+		httpResponse, err := httpClient.Do(request)
+		require.NoError(t, err)
+
+		// Test response
+		require.Equal(t, tt.expectedStatusCode, httpResponse.StatusCode)
+
+		defer func(Body io.ReadCloser) {
+			err := Body.Close()
+			require.NoError(t, err)
```
reinkrul, line 136: no need to check this in the test, I'd think. If it fails, it indicates the test misbehaving (probably due to another assertion failure).

Actually, I don't think I ever close byte streams in tests (since they're so short-lived and then automatically cleaned up). So you could leave it out altogether.

orchestrator/careplanservice/service_test.go
```diff
@@ -46,6 +53,100 @@ func TestService_Proxy(t *testing.T) {
...
+func TestService_Post_Task_Error(t *testing.T) {
+	var tests = []struct {
+		method             string
+		path               string
+		body               string
+		expectedStatusCode int
+		expectedMessage    string
+	}{
+		{
+			"POST",
```
reinkrul, line 71: `http.MethodPost`

orchestrator/careplanservice/service_test.go
```diff
@@ -46,6 +53,100 @@ func TestService_Proxy(t *testing.T) {
...
+			"/cps/Task",
+			"",
+			http.StatusBadRequest,
+			"CarePlanService/CreateTask failed: invalid Task: unexpected end of JSON input",
+		},
+		{
+			"PUT",
```
reinkrul, line 78: `http.MethodPut`

orchestrator/careplanservice/service_test.go
```diff
@@ -46,6 +53,100 @@ func TestService_Proxy(t *testing.T) {
...
+			"/cps/Task/no-such-task",
+			"",
+			http.StatusBadRequest,
+			"CarePlanService/UpdateTask failed: invalid Task: unexpected end of JSON input",
+		},
+		{
+			"PUT",
```
reinkrul, line 85: (etc)

orchestrator/careplanservice/service_test.go
```diff
@@ -46,6 +53,100 @@ func TestService_Proxy(t *testing.T) {
 	require.Equal(t, fhirServerURL.Host, capturedHost)
 }
 
+type OperationOutcomeWithResourceType struct {
+	fhir.OperationOutcome
+	ResourceType *string `bson:"resourceType" json:"resourceType"`
```
reinkrul, line 58: why the `bson` tag?

orchestrator/careplanservice/service.go
```diff
@@ -89,27 +90,24 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
...
 			return
 		}
 	}))
 	mux.HandleFunc("POST /cps/CarePlan", auth.Middleware(authConfig, func(writer http.ResponseWriter, request *http.Request) {
 		err := s.handleCreateCarePlan(writer, request)
 		if err != nil {
-			// TODO: proper OperationOutcome
-			log.Info().Msgf("CarePlanService/CarePlan failed: %v", err)
-			http.Error(writer, "Create CarePlan at CarePlanService failed: "+err.Error(), http.StatusBadRequest)
+			log.Info().Msgf("CarePlanService/CreateCarePlan failed: %v", err)
```
reinkrul, line 109: Can you move these info log messages to the newly introduced function? Since it's such a repeating pattern.

orchestrator/careplanservice/util.go
```diff
@@ -0,0 +1,54 @@
...
+		return
```
reinkrul, line 52: no need to return

reinkrul (changes requested):

orchestrator/careplanservice/service_test.go
tader: That's how all fhir.* structs do this, so I gathered it we also supported that.

reinkrul (approved):

PR #49, opened 2024-09-10
INT-241: Add FHIR proxy for CPC to CPC calls

dwelman:
Adds a FHIR proxy endpoint for method calls from one CPC instance to another.

Still to do: Auth based on provided CareTeam, E2E integ test

### DIFF
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -39,7 +39,7 @@ func New(
 	cpsURL, _ := url.Parse(config.CarePlanService.URL)
 	carePlanServiceClient := fhirclient.New(cpsURL, carePlanServiceHttpClient, coolfhir.Config())
 	fhirClientConfig := coolfhir.Config()
-	transport, _, err := coolfhir.NewFHIRAuthRoundTripper(config.FHIR, fhirClientConfig)
+	transport, _, err := coolfhir.NewAuthRoundTripper(config.FHIR, fhirClientConfig)
 	if err != nil {
 		return nil, err
 	}
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -4,10 +4,12 @@ package careplancontributor
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/SanteonNL/nuts-policy-enforcement-point/middleware"
 	"github.com/SanteonNL/orca/orchestrator/addressing"
 	"github.com/SanteonNL/orca/orchestrator/applaunch/clients"
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
 	"github.com/SanteonNL/orca/orchestrator/user"
+	"github.com/nuts-foundation/go-nuts-client/oauth2"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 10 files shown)


reinkrul (commented):
Can we have an integration test for this? One that we can extend with the requesting-side?

orchestrator/careplancontributor/service.go
```diff
@@ -20,16 +21,33 @@ import (
 const LandingURL = "/contrib/"
 const CarePlanServiceOAuth2Scope = "careplanservice"
 
-func New(config Config, sessionManager *user.SessionManager, carePlanServiceHttpClient *http.Client, didResolver addressing.StaticDIDResolver) *Service {
+func New(config Config, sessionManager *user.SessionManager, carePlanServiceHttpClient *http.Client, didResolver addressing.StaticDIDResolver) (*Service, error) {
+	fhirURL, _ := url.Parse(config.FHIR.BaseURL)
 	cpsURL, _ := url.Parse(config.CarePlanService.URL)
 	carePlanServiceClient := fhirclient.New(cpsURL, carePlanServiceHttpClient, coolfhir.Config())
+	var transport http.RoundTripper
```
reinkrul, line 28: I think we have the same code for the CarePlanService authenticating to its FHIR API. Do we need a shared function there, that works on a common config struct? They are probably the same (it's CPS connecting to local FHIR API, and CPC connecting to local FHIR API, so I'd say the same thing, although the might be separate FHIR APIs).

orchestrator/careplancontributor/service.go
```diff
@@ -50,6 +70,13 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("/contrib/cps/fhir/*", s.withSession(func(writer http.ResponseWriter, request *http.Request, _ *user.SessionData) {
 		carePlanServiceProxy.ServeHTTP(writer, request)
 	}))
+	fhirProxy := coolfhir.NewProxy(log.Logger, s.fhirURL, "/contrib/fhir", s.transport)
+	// TODO: Determine auth. CareTeam can be sent in header but this is not ideal, preferably put it in the access token
```
reinkrul, line 74: can't sent CareTeam in header, as it's not signed (yet).

orchestrator/careplancontributor/service.go
```diff
@@ -50,6 +70,13 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
...
+	// TODO: Modify this and other URLs to /cpc/* in future change
+	mux.HandleFunc("/contrib/fhir/*", s.withSession(func(writer http.ResponseWriter, request *http.Request, _ *user.SessionData) {
+		// TODO: Is this correct?
+		fhirProxy.ServeHTTP(writer, request)
```
reinkrul, line 78: This misses actual authorization (does the resource being requested "belong" to the CarePlan being requested), but it might be a good idea to keep that in an upcoming PR. Then we can simultaneously work on authorization (checking Nuts access token, checking CareTeam) and checking the accessed scope (resource v.s. CarePlan).

orchestrator/careplancontributor/service.go
```diff
@@ -76,6 +103,12 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
 	proxy.ServeHTTP(writer, request)
 }
 
+func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Request, session *user.SessionData) {
```
reinkrul, line 106: unused? Also looks incorrect (the clients.Factories is for the browser app)

orchestrator/careplancontributor/service.go
```diff
@@ -50,6 +70,13 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("/contrib/cps/fhir/*", s.withSession(func(writer http.ResponseWriter, request *http.Request, _ *user.SessionData) {
 		carePlanServiceProxy.ServeHTTP(writer, request)
 	}))
+	fhirProxy := coolfhir.NewProxy(log.Logger, s.fhirURL, "/contrib/fhir", s.transport)
+	// TODO: Determine auth. CareTeam can be sent in header but this is not ideal, preferably put it in the access token
+	// TODO: Modify this and other URLs to /cpc/* in future change
+	mux.HandleFunc("/contrib/fhir/*", s.withSession(func(writer http.ResponseWriter, request *http.Request, _ *user.SessionData) {
```
reinkrul, line 76: `withSession` is for authentication the browser's session at the CPC, so not applicable here. Security comes from authorizing the Nuts access token (`auth.Middleware`)

reinkrul (changes requested):

orchestrator/careplancontributor/service.go
dwelman: Yeah I wanted to get the review out early to confirm I had correctly set up the Proxy, with Auth coming later. Happy for this to be in a future PR

orchestrator/careplancontributor/service.go
dwelman: Thanks, will make the change

orchestrator/lib/coolfhir/azure.go
```diff
@@ -54,3 +55,48 @@ func (a azureTokenSource) Token() (*oauth2.Token, error) {
 		Expiry:      accessToken.ExpiresOn,
 	}, nil
 }
+
+type FHIRRoundTripperConfig struct {
```
dwelman, line 59: Naming is hard, please let me know if there is something better I can name these config structs and the code below

reinkrul: as the package is already "fhir", I'd drop the "FHIR" prefix. Other suggestion: "ClientConfig"?

orchestrator/lib/coolfhir/azure.go
```diff
@@ -54,3 +55,48 @@ func (a azureTokenSource) Token() (*oauth2.Token, error) {
...
+	// BaseURL is the base URL of the FHIR server to connect to.
+	BaseURL string `koanf:"url"`
+	// Auth is the authentication configuration for the FHIR server.
+	Auth FHIRAuthConfig `koanf:"auth"`
+}
+
+type FHIRAuthConfigType string
```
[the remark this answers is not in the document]
dwelman: Sounds good to me, happy with ClientConfig

orchestrator/lib/coolfhir/azure_test.go
```diff
@@ -73,3 +73,59 @@ func Test_azureHttpClient_Do(t *testing.T) {
 		require.Equal(t, "application/fhir+json", capturedHeaders.Get("Content-Type"))
 	})
 }
+
+func Test_NewFHIRAuthRoundTripper(t *testing.T) {
```
reinkrul:
```suggestion
func Test_NewAuthRoundTripper(t *testing.T) {
```

dwelman: Good catch, I used the IDE to rename so it missed this

PR #50, opened 2024-09-12
INT-214: Notify CarePlanContributors when Task or CareTeam is created/changed

reinkrul:
New concepts:
- CSD (Care Service Directory) that can be used to lookup care organizations and their endpoints, that participate in Shared Care Planning
  - It provides a `csd.Directory` as interface, now implemented with Nuts Directory Service
- This CSD is used to lookup FHIR notification endpoints
- `subscriptions.Manager` is used by the CPS to notify CarePlanContributors of changes
- It notifies subscribers (CPCs) through a notification channel (rest-hook).

In scope:
- Notifying CPC when CareTeam or Task is created/updated in CPS
- CPC receives notification and logs it
- CPC secured notification endpoints through Nuts
- CPC publishes notification endpoint in Nuts Discovery Service

Out of scope:
- Provide subscriber with monotonically increasing event number (requires storing Subscription in FHIR store)

Blocked by:
- [x] https://github.com/SanteonNL/orca/pull/52

@bramwesselo I'm not sure we should call it CSD, your throughts?


bramwesselo:
if we'd want to stay close to the [IHE mCSD](https://profiles.ihe.net/ITI/mCSD/index.html), which I would very much like to do, we might say that Orca/Nuts's CSD stands for 'Care Services Discovery' ? (I don't get why IHE puts the 'm' in front of 'mCSD'; they seem to feel some kind of association between FHIR or RESTful and 'mobile')

orchestrator/careplanservice/service.go
```diff
@@ -87,6 +102,15 @@ func (s Service) readRequest(httpRequest *http.Request, target interface{}) erro
 	return json.Unmarshal(data, target)
 }
 
+func (s Service) notifySubscribers(ctx context.Context, resource interface{}) {
+	// Send notification for changed resources
+	notifyCtx, cancel := context.WithTimeout(ctx, subscriberNotificationTimeout)
```
dwelman, line 107: What requirements do we have for failure resolution/retry? If a notification fails to deliver, do we need a DLQ/Redrive or is it fine for the message to not be delivered?

Maybe we could implement a recursive retry with incremental backoff to start with if needed?

reinkrul: Retry is out of scope for this PR, but the general thought is;
- it won't be retried by the CPS
- CPS could perform a heartbeat (it's spec'd in FHIR) to see if the subscriber is available
- CPC will periodically check at the CPS if its subscriber(s) are up-to-date

dwelman: Fine with me, I just wanted to make sure we had it in mind. Thanks!

dwelman (approved):

tader:
I'm still looking at the PR, please allow me some time to review before merging ;)

PR #58, opened 2024-09-17
CPC authorizes requester performing FHIR proxy to CPC

dwelman:
When a request is made from a CPC data requester to a data holder, the requester must first be validated as a member of the careteam with a valid start and end date. Then the request can be proxied

Still TODO: 
* Write E2E integration tests
* Create new Sequence Diagram detailing CPC -> CPC auth


orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+	// Data holder must verify that the requester is part of the CareTeam by checking the URA
+	// Validate by retrieving the CarePlan from CPS, use URA in provided token to validate against CareTeam
+	// CarePlan should be provided in X-SCP-Context header
+	xScpContext := request.Header["X-Scp-Context"]
+	if len(xScpContext) == 0 {
+		// Check if header has different capitalisation, it may be changed by Go canonicalization
+		xScpContext = request.Header["X-SCP-Context"]
+		if len(xScpContext) == 0 {
+			err := errors.New("X-SCP-Context header value must be set")
+			http.Error(writer, err.Error(), http.StatusUnauthorized)
```
reinkrul, line 159: I'd say only return the error, and convert to OperationOutcome where it's called

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+	}
+	carePlanURL := xScpContext[0]
+	if carePlanURL == "" {
+		err := errors.New("X-SCP-Context header value must be set")
+		http.Error(writer, err.Error(), http.StatusUnauthorized)
+		return err
+	}
+	u, err := url.Parse(carePlanURL)
+	if err != nil {
+		return err
```
reinkrul, line 174: We need some kind of validation of the URLs. We do have a CarePlanService URL configured for the CarePlanContributor right? Probably check whether the configured CPS URL starts with the SCP context URL from the header? That mitigates pretty much any "injected" CPS URL attacks.

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
 
+func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Request) error {
+	// Authorize requester before proxying FHIR request
+	// Data holder must verify that the requester is part of the CareTeam by checking the URA
+	// Validate by retrieving the CarePlan from CPS, use URA in provided token to validate against CareTeam
+	// CarePlan should be provided in X-SCP-Context header
+	xScpContext := request.Header["X-Scp-Context"]
+	if len(xScpContext) == 0 {
+		// Check if header has different capitalisation, it may be changed by Go canonicalization
+		xScpContext = request.Header["X-SCP-Context"]
```
reinkrul, line 156: I see Bram changing this header, so make it a constant (also for the error messages) ;-)

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+		return err
+	}
+
+	// Get CarePlan from CPS
+	// TODO: How to go from link in X-SCP-Context to valid resource that can be passed to CPS?
+	carePlanServiceClient := fhirclient.New(s.carePlanServiceURL, s.scpHttpClient, coolfhir.Config())
+	var bundle fhir.Bundle
+	// Use extract CarePlan ID to be used for our query that will get the CarePlan and CareTeam in a bundle
+	carePlanId := strings.TrimPrefix(strings.TrimPrefix(u.Path, "/fhir/CarePlan/"), s.carePlanServiceURL.String())
+	newUrl := fmt.Sprintf("CarePlan\\?_id\\=%s\\&_include\\=CarePlan:care-team", carePlanId)
```
reinkrul, line 183: Why do we need the escaping?

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+	// Get CareTeam from entries
+	for _, entry := range bundle.Entry {
+		err = json.Unmarshal(entry.Resource, &careTeam)
+		// Ignore errors, we are just trying to find the CareTeam
+		if err != nil || len(careTeam.Participant) == 0 {
+			continue
+		}
+		foundCareTeam = true
+		break
+	}
```
reinkrul, line 212: You might be able to replace this with the syntactic sugar of coolfhir:

```
var bundle fhir.Bundle
	var careTeam []fhir.CareTeam
	err := coolfhir.ResourcesInBundle(&bundle, coolfhir.EntryIsOfType("CareTeam"), careTeam)
```

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +145,112 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
 	proxy.ServeHTTP(writer, request)
 }
 
+func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Request) error {
```
reinkrul, line 148: if you just return errors here (instead of writing it to the HTTP response), it makes unit testing it separately much easier

orchestrator/careplancontributor/service.go
dwelman: Yeah that aligns with what I was thinking as well. This was good for initial tests but I agree it should change. Will do

orchestrator/careplancontributor/service.go
dwelman: Sounds good. I have also removed the check for full capitals since the behaviour of the golang http client is to canonicalize the string. Less maintenance headaches later if we need to change the header value

orchestrator/careplancontributor/service.go
dwelman: A bad carry over from when I was testing with cURL. Will remove them

orchestrator/careplancontributor/service.go
dwelman: That does look nicer. Will see if it will work!

orchestrator/careplancontributor/service.go
dwelman: Good idea, that sounds reasonable

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +126,86 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+					}
+					isValidRequester = true
+					break
+				}
+			}
+		}
+	}
+
+	if !isValidRequester {
+		return errors.New("requester does not have access to resource")
```
reinkrul, line 202: Would be cool if we could return something like HTTP 403 in this case. Maybe with a custom error type that lets you specify the HTTP status code?

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +126,86 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+	var bundle fhir.Bundle
+	// Use extract CarePlan ID to be used for our query that will get the CarePlan and CareTeam in a bundle
+	carePlanId := strings.TrimPrefix(strings.TrimPrefix(u.Path, "/fhir/CarePlan/"), s.carePlanServiceURL.String())
+	newUrl := fmt.Sprintf("CarePlan?_id=%s&_include=CarePlan:care-team", carePlanId)
+	err = carePlanServiceClient.Read(newUrl, &bundle)
+	if err != nil {
+		return err
+	}
+	// We are expecting a total of 1 result, with 2 entries (one for CarePlan, one for CareTeam)
+	if *bundle.Total == 0 {
```
reinkrul, line 163: wouldn't checking `len(bundle.Entry)` be safer? (and constant with the check below)

And I think if we leave out the 2 checks , the following check would stil fail right:

```
	if len(careTeams) == 0 {
		return errors.New("CareTeam not found in bundle")
	}
```

so do we need them? Or do we need to reassure the CarePlan exists (although I don't really see that being explciti here).

dwelman: I can see an error for the CarePlan not existing being useful as an error to return to the user so I will do an explicit check for that, but otherwise yes I think the other checks can be removed to streamline the validation.

Not sure why I checked bundle.Total... Definitely not consistent with the rest of the validation

orchestrator/careplancontributor/service.go
dwelman: That sounds like a good addition, will add it

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +126,83 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
 	proxy.ServeHTTP(writer, request)
 }
 
+func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Request) error {
+	// Authorize requester before proxying FHIR request
+	// Data holder must verify that the requester is part of the CareTeam by checking the URA
+	// Validate by retrieving the CarePlan from CPS, use URA in provided token to validate against CareTeam
+	// CarePlan should be provided in X-Scp-Context header
+	carePlanURLValue := request.Header[carePlanURLHeaderKey]
+	if len(carePlanURLValue) == 0 {
```
reinkrul, line 135: missing test coverage, collapse these 2 checks (len == 0, len == 1), having them separate doesn't really add value i.m.o

orchestrator/lib/coolfhir/util.go
```diff
@@ -64,6 +65,57 @@ func ValidateLogicalReference(reference *fhir.Reference, expectedType string, ex
 	return nil
 }
 
+// ValidateCareTeamParticipantPeriod validates that a CareTeamParticipant has a start date, and that the start date is in the past
+// end date is not required, but if present it will validate that it is in the future
+func ValidateCareTeamParticipantPeriod(participant fhir.CareTeamParticipant) error {
```
reinkrul, line 70: provide `now` in function signatures, eases testing an makes function applicable in more instances

reinkrul (changes requested):

orchestrator/careplancontributor/service.go
```diff
@@ -117,6 +126,83 @@ func (s Service) handleProxyToEPD(writer http.ResponseWriter, request *http.Requ
...
+	}
+	u, err := url.Parse(carePlanURL)
+	if err != nil {
+		return err
+	}
+
+	carePlanServiceClient := fhirclient.New(s.carePlanServiceURL, s.scpHttpClient, coolfhir.Config())
+	var bundle fhir.Bundle
+	// Use extract CarePlan ID to be used for our query that will get the CarePlan and CareTeam in a bundle
+	carePlanId := strings.TrimPrefix(strings.TrimPrefix(u.Path, "/fhir/CarePlan/"), s.carePlanServiceURL.String())
```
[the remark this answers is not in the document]
dwelman: Yes, good idea

orchestrator/careplancontributor/service.go
dwelman: Yeah, fair point

orchestrator/lib/coolfhir/util.go
dwelman: Yes that's a good call, will add it

orchestrator/careplancontributor/integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+		if err := container.Terminate(ctx); err != nil {
+			panic(err)
+		}
+	})
+	endpoint, err := container.Endpoint(ctx, "http")
+	require.NoError(t, err)
+	u, err := url.Parse(endpoint)
+	require.NoError(t, err)
+	return u.JoinPath("fhir")
+}
```
reinkrul, line 213: I think this is duplicate now, maybe we move it to somewhere, where both integration tests can use it?

orchestrator/careplancontributor/service.go
```diff
@@ -149,13 +152,16 @@ func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Req
 	if err != nil {
 		return err
 	}
+	// Verify that the u.Path refers to a careplan
+	if !strings.HasPrefix(u.Path, "/cps/CarePlan/") {
+		return errors.New("header does not include CarePlan")
```
reinkrul, line 157:
```suggestion
		return errors.New("specified SCP context header does not refer to a CarePlan")
```

orchestrator/careplancontributor/service.go
```diff
@@ -182,15 +188,21 @@ func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Req
...
 				}
+				if isValidRequester {
+					break
+				}
+			}
+			if isValidRequester {
+				break
 			}
 		}
 	}
```
reinkrul, line 208: the multiple if-breaks make this hard to grasp, my solution would be a separate func where you can just `return` from, or add a GOTO-ish label to the outer for-loop (and use `break outer/(or something`)

reinkrul (changes requested):

orchestrator/careplancontributor/integration_test.go
dwelman: Sounds good, will do

orchestrator/careplancontributor/service.go
dwelman: Yeah I am not happy with this either, I will clean it up. Thanks

orchestrator/careplancontributor/service.go
```diff
@@ -184,35 +184,38 @@ func (s Service) handleProxyToFHIR(writer http.ResponseWriter, request *http.Req
...
 		for _, participant := range careTeam.Participant {
 			for _, identifier := range principal.Organization.Identifier {
 				if coolfhir.IdentifierEquals(participant.OnBehalfOf.Identifier, &identifier) {
 					// Member must have start date, this date must be in the past, and if there is an end date then it must be in the future
-					err = coolfhir.ValidateCareTeamParticipantPeriod(participant, time.Now())
+					err := coolfhir.ValidateCareTeamParticipantPeriod(participant, time.Now())
 					if err != nil {
-						return err
+						return false, err
 					}
```
reinkrul, line 212: There's an edge case (although currently not possible now); if the principal's org has multiple identifiers, of which multiple occur in the careTeam, if the first match has an expired period, it invalidates the membership. But, a there could be a subsequent membership that's not expired. Although it would be an extreme edge case, I think it would simplify things if you just return `true` when there's a matching identifier with a non-expired period.

I think this roots from the fact that `ValidateCareTeamParticipantPeriod` should return a bool (whether or not expired) AND an error (invalid object). In case of an error, you can just log it (WARN) and ignore the entry, and try the next identifier.

reinkrul (approved):
Almost there, last thing.

orchestrator/careplancontributor/service.go
dwelman: That sounds reasonable, I hadn't considered that edge case so thanks for bringing it to my attention!

orchestrator/careplancontributor/service.go
```diff
@@ -206,11 +206,13 @@ func validateRequester(careTeams []fhir.CareTeam, principal auth.Principal) (boo
 			for _, identifier := range principal.Organization.Identifier {
 				if coolfhir.IdentifierEquals(participant.OnBehalfOf.Identifier, &identifier) {
 					// Member must have start date, this date must be in the past, and if there is an end date then it must be in the future
-					err := coolfhir.ValidateCareTeamParticipantPeriod(participant, time.Now())
+					ok, err := coolfhir.ValidateCareTeamParticipantPeriod(participant, time.Now())
 					if err != nil {
-						return false, err
+						log.Warn().Msg("error while validating CareTeam Participant period: " + err.Error())
```
reinkrul, line 211: There's a function in the logger to add the error as field, I think it's something like `Err(e)`
```suggestion
						log.Warn().Err(err).Msg("invalid CareTeam Participant period")
```

reinkrul (approved):

dwelman:
When a request is made from a CPC data requester to a data holder, the requester must first be validated as a member of the careteam with a valid start and end date. Then the request can be proxied

PR #64, opened 2024-09-27
Orchestrator: embed questionnaires in FS instead of hardcode

reinkrul:
First step in implementing a flexible Task Engine; moving the hardcoded questionnaires to embedded assets. Also starts introducing a "task engine" layer with the concept of a "workflow" (not sure if that's the right term), which is extendable for loading questionnaires from other sources (and handle things like caching).

Also allows easier update of the questionnaire when it changes during development.


orchestrator/careplanservice/taskengine/assets/assets.go
JorisHeadease, line 1: Ah, so this is why I couldn't get my json files loaded from the filesystem!

JorisHeadease (approved):
Nice setup @reinkrul!

PR #67, opened 2024-10-01
Change CPC base URL from /contrib to /cpc for consistency with CPS

dwelman:


reinkrul (approved):
LGTM

dwelman:
Having issues running the FE locally, might have missed changing something. Will update the PR once I have it working again

PR #68, opened 2024-10-02
CPS: Rework Bundle processing to dispatch to individual FHIR operations

reinkrul:
The general idea is as follows;
- There are handlers for "simple" FHIR operations, create/update Task, create/update X, or Y (GET operations are a separate story).
- These handlers can be executed as-is (for e.g. `POST /Task`), or as part of a Bundle.
- This works as follows:
  - each handler is called, and passed a "transaction" Bundle. The handlers add FHIR operations (e.g. create Task, update CareTeam) to the transaction as they need. Each handler returns a result processor. This is done by `proceedTransaction`.
  - The Bundle handler executes the transaction, and passes the result to each of the result processors. What comes out is compiled into a Bundle that is returned to the client. This is done by `commitTransaction`.
- For simple FHIR operations, the sole result is returned as-is.
- Notifications are collected and sent to subscribers during `commitTransaction`.

TODO:
- [x] Rewrite Update Task and Create CarePlan to follow the new pattern
- [x] Write unit tests for the Bundle dispatch logic


orchestrator/careplanservice/service.go
```diff
@@ -60,51 +67,204 @@ type Service struct {
...
+		notificationResources = append(notificationResources, currNotificationResources...)
+	}
+
+	for _, notificationResource := range notificationResources {
+		s.notifySubscribers(request.Context(), notificationResource)
+	}
+	return &resultBundle, nil
+}
+
+func (s *Service) proceedTransaction(writer http.ResponseWriter, request *http.Request, resourcePath string, tx *coolfhir.TransactionBuilder) (FHIRHandlerResult, error) {
```
dwelman, line 139: Nitpick: It's a bit difficult to understand what this method does by name, maybe something like `ProgressTransaction` or `handleTransaction` would be clearer?

dwelman (approved):
From my current understanding this looks good and I like the testing you have added, that builds a lot of confidence. I will continue to understand how this works once I start making my own changes going forward but happy for now and glad we addressed concerns in the huddle. Minor nitpick but otherwise looks good

orchestrator/careplanservice/service.go
reinkrul: changes it to `handleTransactionEntry` and added more comments

PR #69, opened 2024-10-03
In-process notification when CPC and CPS run on the same SCP-node

reinkrul:
When CPS and CPC run on the same node, CPS can just notify the CPC using an in-process struct. In this case, I decided to have a simple, global "event handler" for handling subscription notifications.

This makes the system more robust, as there is less moving parts that could break (e.g. access token negotation, validation/authorization, etc).

"The same node" is checked if the local SCP-node "owns" the FHIR identifier that the notification is sent to, by consulting the SCP-nodes profile. Currenly, as we only support the "Dutch Nuts profile", it means it derives the FHIR identifiers of the local node from the NutsUraCredentials in the Nuts node's wallet.

TODO:
- [x] Tests


dwelman (approved):
Happy with these changes so far, would be good to see tests for the csd.go changes to verify the fetching of cache after expiry, but I do see you have tests in your TODO so that is fine

PR #71, opened 2024-10-04
Support unhandled FHIR operations within Bundle & expand e2e-test

reinkrul:
The recent Bundle refactor broke support for FHIR operations within Bundles that we don't explicitly handle ("unmanaged"). Although everything needs to be managed/explicitly handled in the near future, not everything is (e.g. creating Questionnaire/-Response). This worked when creating/updating it as single resource, but not as a part of a Bundle. This PR fixes this, with a simplification of abstracting the FHIR operation (earlier HTTP request) with a `FHIRHandlerRequest`. This also allows to more easily support FHIR headers in future (e.g. ETag).

Also the local identifier handling was broken (you create 2 resources in 1 Bundle that refer to each other, having the FHIR server resolve and replace the references using `FullUrl`), which is now also fixed again.

Changes:
- Fix support for unhandled FHIR operations within Bundles
- Expand e2e tests to cover Task Filler Engine and responding to Questionnaires


dwelman (approved):
Nice work with the bundle refactor! Lots of upfront work but it'll be good for the long term sustainability of the system

PR #74, opened 2024-10-10
[INT-204] Validate CRUD methods

dwelman:
Add validation to CRUD methods for CarePlan, CareTeam, and Task, only allowing valid participants of a CarePlan to access resources related to that CarePlan.

This will be a large change so I am going to commit in stages:

- [x] Refactor Task creation to also create a CarePlan and CareTeam if one is not provided
- [x] Validate Task creation
- [x] Validate Task update
- [x] Validate Get request before Proxy
- [ ] Validate requirements for CarePlan Creation
- [x] Unit tests for new handlers 
- [x] Refactor Get and Search handlers


orchestrator/careplanservice/handle_createtask.go
```diff
@@ -30,17 +34,74 @@ func (s *Service) handleCreateTask(httpRequest *http.Request, tx *coolfhir.Trans
...
+
+		task.BasedOn = []fhir.Reference{
+			{
+				Type:      to.Ptr("CarePlan"),
+				Reference: to.Ptr("CarePlan/" + *carePlan.Id),
+			},
+		}
+	} else {
+		// Adding a task to an existing CarePlan
+		carePlanRef, err = basedOn(task)
```
reinkrul, line 81: We should be starting to write unit tests, as we're working towards production-ready code (and relying on slow integration tests will ultimately slow down development). Should be easy, since the primary output of these handlers is now the Bundles entries they add to the transaction (`tx`). See `handle_createcareplan_test.go` for an example (`TestService_handleCreateCarePlan`)

orchestrator/careplanservice/service.go
```diff
@@ -259,8 +261,6 @@ func (s *Service) defaultHandlerProvider(method string, resourcePath string) fun
 		switch getResourceType(resourcePath) {
 		case "Task":
 			return s.handleCreateTask
-		case "CarePlan":
-			return s.handleCreateCarePlan
```
reinkrul, line 263: if this is now unused, remove the code

orchestrator/careplanservice/handle_get.go
```diff
@@ -0,0 +1,191 @@
...
+	fhirclient "github.com/SanteonNL/go-fhir-client"
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
+	"github.com/SanteonNL/orca/orchestrator/lib/coolfhir"
+	"github.com/rs/zerolog/log"
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+	"net/http"
+	"strings"
+)
+
+func (s *Service) handleGet(httpResponse http.ResponseWriter, request *http.Request) error {
```
reinkrul, line 15: I'd split this in handler functions, just like we do for the write operations? Why diverge?

orchestrator/careplanservice/handle_get.go
```diff
@@ -0,0 +1,191 @@
...
+	queryParams := request.URL.Query()
+	urlParts := strings.Split(strings.TrimPrefix(request.URL.Path, basePath+"/"), "/")
+	switch urlParts[0] {
+	case "CarePlan":
```
reinkrul, line 19: This is actually a search, might be more clear if you have separate routing for the two. E.g.:

```go
// Handle reading a specific resource instance
mux.HandleFunc("GET "+basePath+"/{type}/{id}", ...)
// Handle search
mux.HandleFunc("GET "+basePath+"/{type}", ...)
```

or specific for the resource type;

```go
```go
// Handle reading a specific resource instance
mux.HandleFunc("GET "+basePath+"/CarePlan/{id}", ...)
// Handle search
mux.HandleFunc("GET "+basePath+"/CarePlan", ...)
```

The latter is more like how the create funcs are working
```

File: orchestrator/careplanservice/handle_get.go [hunk 5]
```diff
@@ -0,0 +1,191 @@
...
+				return coolfhir.NewErrorWithCode("CareTeam not found in bundle", http.StatusNotFound)
+			}
+
+			err = validatePrincipalInCareTeams(request.Context(), careTeams)
+			if err != nil {
+				return err
+			}
+
+			// Valid requester, now perform the bundle request again with all the params
+			params := []fhirclient.Option{}
```
reinkrul, line 47:
```suggestion
			// TODO: Check if all parameters are allowed
			params := []fhirclient.Option{}
```
@bramwesselo ?

orchestrator/careplanservice/handle_get.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (s *Service) handleGet(httpResponse http.ResponseWriter, request *http.Request) error {
+	queryParams := request.URL.Query()
+	urlParts := strings.Split(strings.TrimPrefix(request.URL.Path, basePath+"/"), "/")
+	switch urlParts[0] {
+	case "CarePlan":
+		// fetch CarePlan + CareTeam, validate requester is participant of CareTeam
+		// if there are query params present, we need to handle the request a bit differently
+		if len(queryParams) != 0 {
+			if queryParams.Get("_id") == "" {
```
reinkrul, line 23: why would the caller be using search if they have the ID of the CarePlan they want to request? SInce then they could just be getting the CarePlan with this ID? (`CarePlan/(id)`), then they won't get a bundle in return (sounds easier)

orchestrator/careplanservice/handle_get.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			return err
+		}
+
+		b, err := json.Marshal(careTeam)
+		_, err = httpResponse.Write(b)
+		if err != nil {
+			return err
+		}
+		// TODO: Do we support additional info in the URL i.e. extra params
```
reinkrul, line 108: Caller won't get the headers returned by the FHIR API. You can capture these with `fhirclient.ResponseHeaders(headers)` option for `Read()`, then set them on the HTTP response;
```
		for key, value := range headers.Header {
			httpResponse.Header()[key] = value
		}
```

reinkrul: (I took this from `lib/coolfhir/bundle.go:195`)

reinkrul: (also applies to the other handlers in this file)

orchestrator/careplanservice/handle_get.go
```diff
@@ -0,0 +1,191 @@
...
+func validatePrincipalInCareTeams(ctx context.Context, careTeams []fhir.CareTeam) error {
+	// Verify requester is in CareTeams
+	principal, err := auth.PrincipalFromContext(ctx)
+	if err != nil {
+		return err
+	}
+	participant := coolfhir.FindMatchingParticipantInCareTeam(careTeams, principal.Organization.Identifier)
+	if participant == nil {
+		return &coolfhir.ErrorWithCode{
+			Message:    "Participant is not part of CarePlan",
```
reinkrul, line 186: CareTeam

orchestrator/careplanservice/handle_updatetask.go
```diff
@@ -32,6 +32,16 @@ func (s *Service) handleUpdateTask(httpRequest *http.Request, tx *coolfhir.Trans
 		return nil, fmt.Errorf("failed to read Task: %w", err)
 	}
 
+	// Validate fields on updated Task
+	// if Intent is not set on the updated task then just get it from the existing task, otherwise run usual validation rules
+	if task.Intent == "" {
+		task.Intent = taskExisting.Intent
```
reinkrul, line 38: why default, and not return an error?

reinkrul (changes requested):

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -30,17 +34,74 @@ func (s *Service) handleCreateTask(httpRequest *http.Request, tx *coolfhir.Trans
...
+
+		bundle, err := s.commitTransaction(newHttpRequest, newTx, []FHIRHandlerResult{res})
+		if err != nil {
+			return nil, err
+		}
+
+		err = coolfhir.ResourceInBundle(bundle, coolfhir.EntryIsOfType("CarePlan"), &carePlan)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
dwelman: I figured this might cause problems with the transaction. Sounds good, thanks, will look into it

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -30,17 +34,74 @@ func (s *Service) handleCreateTask(httpRequest *http.Request, tx *coolfhir.Trans
...
 	}
-	// TODO: Manage time-outs properly
-	var carePlan fhir.CarePlan
-	if err := s.fhirClient.Read(*carePlanRef, &carePlan); err != nil {
-		return nil, fmt.Errorf("failed to read CarePlan: %w", err)
+	// Resolve the CarePlan
+	if len(task.BasedOn) == 0 {
+		// The CarePlan does not exist, a CarePlan and CareTeam will be created and the requester will be added as a member
+		cp := fhir.CarePlan{}
+
```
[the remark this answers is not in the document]
dwelman: Bram and I discussed this briedly during our sync, there were still some open questions about how to handle this auth and if it is in scope for now. Would be good to further discuss

orchestrator/careplanservice/handle_createtask.go
dwelman: Yes that's a fair point! Unit tests were still on the way but I wanted to prioritise on getting the flow working with the integ tests to also start with the review

orchestrator/careplanservice/handle_get.go
dwelman: This actually came up due to a search we are doing in CPC:
```
carePlanId := strings.TrimPrefix(strings.TrimPrefix(u.Path, "/cps/CarePlan/"), s.carePlanServiceURL.String())
	err = carePlanServiceClient.Read("CarePlan", &bundle, fhirclient.QueryParam("_id", carePlanId), fhirclient.QueryParam("_include", "CarePlan:care-team"))
	if err != nil {
		return err
	}
```

The CareTeam is required to check if the participant has a valid period

orchestrator/careplanservice/handle_get.go
dwelman: Looks good, thanks!

orchestrator/careplanservice/handle_createtask.go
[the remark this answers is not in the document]
bramwesselo: How about enabling this on a private endpoint? Or autorize/require the creation of a Task without reference to a Careplan only permitted by the organization/URA that is running the CPS?

orchestrator/careplanservice/handle_updatetask.go
dwelman: I figured it might be annoying if you didn't set the field but you didn't want it to change, that being said though I realise this is a PUT and not a PATCH and my reasoning doesn't make sense in that case. I will remove this and let the existing validation handle it

orchestrator/careplanservice/handle_get.go
bramwesselo: well, I'm trying to make sense of this code. I see that there's a fhir-client-repo, some coolfhir-lib and a zorgbijjou-fhir-model-fork; it's a lot to get my head around...

dwelman: I have pushed this out in distinct commits as best I could, might be easier to review changes per commit

orchestrator/careplanservice/handle_get.go
dwelman: I think I can modify this so we don't have to worry about _id being passed in, will test my idea out

orchestrator/careplanservice/handle_createtask.go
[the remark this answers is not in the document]
reinkrul: Options (from most open to most closed); 
- keep it "open" for every care organization (having an URA credential)
- only allow configured URAs to create CarePlans
- have it as private endpoint for the local EHR

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -20,11 +20,12 @@ import (
 
 func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerRequest, tx *coolfhir.TransactionBuilder) (FHIRHandlerResult, error) {
 	log.Info().Msg("Creating Task")
-	// TODO: Check only allowed fields are set, or only the allowed values (INT-204)?
+	log.Info().Msgf("%s", request.RequestUrl)
 	var task fhir.Task
 	if err := json.Unmarshal(request.ResourceData, &task); err != nil {
 		return nil, fmt.Errorf("invalid %T: %w", task, err)
 	}
+	log.Info().Msgf("%v", task)
```
reinkrul, line 28: leftover debug logging?

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -33,47 +34,54 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
-		if err != nil {
-			return nil, err
-		}
-
-		err = coolfhir.ResourceInBundle(bundle, coolfhir.EntryIsOfType("CarePlan"), &carePlan)
-		if err != nil {
-			return nil, err
-		}
-
+		// Create a new CarePlan (which will also create a new CareTeam) based on the Task reference
```
reinkrul, line 51: this looks much better and convoluted

orchestrator/careplanservice/handle_get.go
```diff
@@ -15,17 +15,27 @@ import (
...
+
 	switch urlParts[0] {
 	case "CarePlan":
 		// fetch CarePlan + CareTeam, validate requester is participant of CareTeam
 		// if there are query params present, we need to handle the request a bit differently
 		if len(queryParams) != 0 {
-			if queryParams.Get("_id") == "" {
-				return coolfhir.NewErrorWithCode("Query parameter _id is not present, can't find CarePlan", http.StatusBadRequest)
+			// TODO: Check if all parameters are allowed
+			// TODO: Is it acceptable to return the CareTeam in the bundle if it is not requested?
```
reinkrul, line 26: If it's a Search operation, a Bundle is expected, then it might be semantically acceptable. But I'd think it would be weird and don't understand why we would need to.

orchestrator/careplanservice/handle_get.go
```diff
@@ -81,8 +85,13 @@ func (s *Service) handleGet(httpResponse http.ResponseWriter, request *http.Requ
 		if err != nil {
 			return err
 		}
+
+		for key, value := range headers.Header {
+			httpResponse.Header()[key] = value
+		}
 		// TODO: Do we support additional info in the URL i.e. extra params
 	case "CareTeam":
+		// TODO: Support bundle search for CareTeam, once logic for CarePlan is agreed upon
```
reinkrul, line 94: let's discuss this at the office

orchestrator/careplanservice/handle_get.go
dwelman: An early draught just to get the flow working, these methods will be split into their own handlers in the next revision

orchestrator/careplanservice/handle_get.go
dwelman: I have implemented this for the next revision

orchestrator/careplanservice/handle_createtask.go
dwelman: Yes, good catch. It has been removed

orchestrator/careplanservice/handle_createtask.go
dwelman: Hopefully you mean "much better and *less* convoluted" :P

orchestrator/careplanservice/handle_get.go
dwelman: Discussed with Bram, should be fine

orchestrator/careplanservice/handle_gettask.go
```diff
@@ -0,0 +1,98 @@
...
+// Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
+func (s *Service) handleSearchTask(ctx context.Context, queryParams url.Values, headers *fhirclient.Headers) (*fhir.Bundle, error) {
+	params := []fhirclient.Option{}
+	for k, v := range queryParams {
+		params = append(params, fhirclient.QueryParam(k, v[0]))
+	}
+
+	params = append(params, fhirclient.ResponseHeaders(headers))
+	var bundle fhir.Bundle
+	err := s.fhirClient.Read("Task", &bundle, params...)
```
reinkrul, line 60: wonder if this won't become too slow without search narrowing (@bramwesselo )

orchestrator/careplanservice/service.go
```diff
@@ -132,11 +132,21 @@ func (s *Service) RegisterHandlers(mux *http.ServeMux) {
...
+		resourceType := request.PathValue("type")
+		resourceId := request.PathValue("id")
+		s.handleGet(request, httpResponse, resourceId, resourceType, "CarePlanService/Get"+resourceType)
 	}))
+	// Handle search
+	mux.HandleFunc("GET "+basePath+"/{type}", s.profile.Authenticator(baseUrl, func(httpResponse http.ResponseWriter, request *http.Request) {
+		resourceType := request.PathValue("type")
+		s.handleSearch(request, httpResponse, resourceType, "CarePlanService/Get"+resourceType)
+	}))
+	//mux.HandleFunc("GET "+basePath+"/{resourcePath...}", s.profile.Authenticator(baseUrl, func(httpResponse http.ResponseWriter, request *http.Request) {
```
reinkrul, line 146: remove

orchestrator/careplanservice/service.go
```diff
@@ -237,6 +247,76 @@ func (s *Service) handleCreateOrUpdate(httpRequest *http.Request, httpResponse h
...
+	b, err := json.Marshal(resource)
+	_, err = httpResponse.Write(b)
+	if err != nil {
+		coolfhir.WriteOperationOutcomeFromError(err, operationName, httpResponse)
+		return
+	}
+
+	for key, value := range headers.Header {
+		httpResponse.Header()[key] = value
+	}
```
reinkrul, line 281: wrong order, set headers first, then write response body

orchestrator/careplanservice/service.go
```diff
@@ -237,6 +247,76 @@ func (s *Service) handleCreateOrUpdate(httpRequest *http.Request, httpResponse h
...
+	b, err := json.Marshal(bundle)
+	_, err = httpResponse.Write(b)
+	if err != nil {
+		coolfhir.WriteOperationOutcomeFromError(err, operationName, httpResponse)
+		return
+	}
+
+	for key, value := range headers.Header {
+		httpResponse.Header()[key] = value
+	}
```
reinkrul, line 316: headers first, then response code, then response body

orchestrator/careplanservice/util.go
```diff
@@ -40,3 +43,37 @@ func (s *Service) writeOperationOutcomeFromError(err error, desc string, httpRes
...
+	// Verify requester is in CareTeams
+	principal, err := auth.PrincipalFromContext(ctx)
+	if err != nil {
+		return err
+	}
+	participant := coolfhir.FindMatchingParticipantInCareTeam(careTeams, principal.Organization.Identifier)
+	if participant == nil {
+		return &coolfhir.ErrorWithCode{
+			Message:    "Participant is not part of CareTeam",
+			StatusCode: http.StatusUnauthorized,
```
reinkrul, line 75: 403 Forbidden might be better

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"net/url"
+)
+
+// handleGetCarePlan fetches the requested CarePlan and validates if the requester has access to the resource (is a participant of one of the CareTeams of the care plan)
+// if the requester is valid, return the CarePlan, else return an error
+// Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
+func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhirclient.Headers) (*fhir.CarePlan, error) {
+	// fetch CarePlan + CareTeam, validate requester is participant of CareTeam
+	carePlan, careTeams, headers, err := s.getCarePlanAndCareTeams("CarePlan/" + id)
```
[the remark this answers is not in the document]
dwelman: The headers are passed in to handleGetCarePlan by reference, would you like me to rewrite this to return the headers instead?

orchestrator/careplanservice/handle_createtask.go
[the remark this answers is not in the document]
bramwesselo: I thought we settled on option: "only allow configured URAs to create CarePlans". We could start by allowing our 'own' URA by default if that helps?

reinkrul (approved):

orchestrator/careplanservice/handle_gettask.go
dwelman: I'll create a separate task for this so this code can be merged so long

PR #81, opened 2024-10-23
Int 371 - BUG | CPS Bundle handling returns unexpected results (SAZ-test)

rolandgroen:
- Changed the parsing of the reference, used `url.Parse` to handle the URL and extract the path. 
- Added test for the condition.


orchestrator/careplanservice/service.go
```diff
@@ -351,16 +351,25 @@ func (s *Service) handleBundle(httpRequest *http.Request, httpResponse http.Resp
...
+		}
+		if requestUrl.IsAbs() {
+			coolfhir.WriteOperationOutcomeFromError(fmt.Errorf("bundle.entry[%d].request.url (entry #) must be a relative URL", entryIdx), op, httpResponse)
+			return
+		}
+		resourcePath := requestUrl.Path
 		resourcePathParts := strings.Split(resourcePath, "/")
 		if entry.Request == nil || len(resourcePathParts) > 2 {
-			coolfhir.WriteOperationOutcomeFromError(fmt.Errorf("bundle.entry[%d].request.url (entry #) must be a relative URL", entryIdx), op, httpResponse)
+			coolfhir.WriteOperationOutcomeFromError(fmt.Errorf("bundle.entry[%d].request.url (entry #) has too many paths", entryIdx), op, httpResponse)
```
bramwesselo, line 366: The SAZ bundle (with url Patient?identifier=http......//...//../|111222333) will still crash now, or? (i.e. this does solve the problem....?)

rolandgroen: Yes, because all the stuff behind the ? is not part of the path:
```
requestUrl, err := url.Parse(entry.Request.Url)
```
Will parse the url `Patient?identifier=http......//...//../|111222333`
```
resourcePath := requestUrl.Path
```
Will result in `Patient`
```
resourcePathParts := strings.Split(resourcePath, "/")
```
Wil get resourcePathParts[0] only

dwelman (approved):

PR #82, opened 2024-10-24
No longer send a CarePlan from CPC to CPS

JorisHeadease:
According to the spec, the CPS should create the CP - when there is no Task.basedOn on the Task created by the placer. Removed this from the frontend. The backend has already been modified.


JorisHeadease:
The code seems to run into some bugs still, but they seem unrelated to this PR:

1. When the CPC requests all the `CarePlans` for a `Patient` based on the BSN, the service now returns a 404 with the message "CarePlanService/GetCarePlan failed: CareTeam not found in bundle"
2. When the CPC sends the `Task` to the CPS, the service returns a 400 with the message "CarePlanService/CreateTask failed: failed to load local identities: list credentials non-OK HTTP response (status=400 Bad Request): HTTP 400 - GetCredentialsInWallet failed - invalid holder DID: invalid DID"

When I test this PR against a local fhir instance, the `Task` is created without the `basedOn` reference just fine.

dwelman (approved):

PR #83, opened 2024-10-25
[INT-309] Move taskfiller from CPS to CPC

dwelman:
Moves the taskfiller to CPC, triggered by the notification mechanism in CPS.

Currently only supports creation of task, as was supported in CPS


orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
...
+		}
+	}
+
+	focusReference, err := notification.GetFocus()
+	if err != nil {
+		return err
+	}
+	if focusReference == nil {
+		return &coolfhir.ErrorWithCode{
+			Message:    "Notification focus now found",
```
JorisHeadease, line 262:
```suggestion
			Message:    "Notification focus not found",
```

orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
...
+			StatusCode: http.StatusInternalServerError,
```
JorisHeadease, line 263: Probably more of a 422 response code?

dwelman:
The E2E test is currently failing due to the notifications not being sent, something to do with the discovery service. I am not familiar with how this works so if I can't figure it out today I will consult @reinkrul next week

orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
...
+	switch *focusReference.Type {
+	case "Task":
+		// Get task
+		var task fhir.Task
+		err = s.carePlanServiceClient.Read(*focusReference.Reference, &task)
+		if err != nil {
+			return err
+		}
+
+		// TODO: How to differentiate between create and update? (Currently we only use Create in CPS. There is code for Update but nothing calls it)
```
JorisHeadease, line 276: I am not sure if something is built into the notification solution we use, we need to check that first. If not:

Every task has a version in the `Task.meta.versionId`. Mostly, this is an incremental integer, but FHIR clearly states that the code must not assume this to be an incremental integer.

> Changes each time the content of the resource changes. Can be referenced in a [resource reference](https://www.hl7.org/fhir/r4/references.html#Resource). Can be used to ensure that updates are based on the latest version of the resource.
The version can be globally unique, or scoped by the Logical Id of the resource. Version identifiers are generally either a serially incrementing id scoped by the logical id, or a uuid, though neither of these approaches is required. There is no fixed order for version ids - clients cannot assume that a versionId that comes after another one either numerically or alphabetically represents a later version. The same versionId can never be used for more than one version of the same resource.
On the RESTful API: On receiving a [write operation](https://www.hl7.org/fhir/r4/http.html#write), the server SHALL update this item to the current value, or remove it. Note that servers SHOULD support versions, but some are unable to

So, if we cannot rely on the versionId, a safer (but less performant) option is to:
```
GET [baseURL]/Task/[id]/_history?_summary=count
```
This will provide the count of versions for a given Task.

orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
...
+		err = s.handleTaskFillerCreate(&task)
```
JorisHeadease, line 277: Shouldn't this specific CPC node verify that this Task is created for them (via the `Task.owner`)? 

I'm unsure how the notification logic works now, but I can imagine the placer's CPC is also informed when they create a Task. Could also be added inside the `handleTaskFillerCreate()` call

orchestrator/careplancontributor/service.go
dwelman: I am hesitant to add this change in this PR considering the task was to simply move the functionality and I would want to modify it as little as needed. That being said, if it is required for feature parity (considering this is now in CPC instead of CPS and is now an async process), then it would need to be implemented.

I am not fully aware of the impact so I will bring it up during the team discussions on Monday

orchestrator/careplancontributor/service.go
dwelman: Hmm, that's a good point. We do have validation for creating a task and the notification is sent based on the Task.Owner so I think we should be fine, but it might be a nice extra layer of validation. Will discuss this with Rein on Monday and see if it will be required

orchestrator/careplancontributor/service.go
JorisHeadease: For sure, it was more of an informative comment answering the TODO, indeed out of scope for this PR!

reinkrul:
> The E2E test is currently failing due to the notifications not being sent, something to do with the discovery service. I am not familiar with how this works so if I can't figure it out today I will consult @reinkrul next week

Fixed by https://github.com/SanteonNL/orca/pull/84

orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
 	}
 }
 
-func (s Service) handleNotification(_ any) error {
-	log.Info().Msg("TODO: Handle received notification")
+func (s Service) handleNotification(resource any) error {
+	notification := resource.(*coolfhir.SubscriptionNotification)
+	if notification == nil {
```
reinkrul, line 249: this will only be `true` if `notification` is actually `nil`, if it's something other, the cast will panic.

orchestrator/careplancontributor/service.go
```diff
@@ -259,7 +244,43 @@ func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter
...
+	if err != nil {
+		return err
+	}
+	if focusReference == nil {
+		return &coolfhir.ErrorWithCode{
+			Message:    "Notification focus not found",
+			StatusCode: http.StatusUnprocessableEntity,
+		}
+	}
+
```
reinkrul, line 266: maybe some debug/info logging concerning that we have received a notification?

orchestrator/careplancontributor/service.go
reinkrul: any SCP-node can send any notification, meaning they're from an untrusted source. So they need to be validated.

> We do have validation for creating a task and the notification is sent based on the Task.Owner

You refer to the checks done by the CPS here, right? Since the CPS is external (or might be), we can't rely on that, thus `handleTaskFillerCreate` should validate it partakes in the `Task`. Furthermore; it should check whether it's actually the filler of the Task (because if it's the placer itself, which might/is also notified, it obviously shouldn't start responding to the Task it created itself).

reinkrul (changes requested):

orchestrator/careplancontributor/service.go
```diff
@@ -43,28 +44,33 @@ func New(
...
+		orcaPublicURL:         orcaPublicURL,
+		carePlanServiceURL:    cpsURL,
+		carePlanServiceClient: carePlanServiceClient,
+		SessionManager:        sessionManager,
+		scpHttpClient:         httpClient,
+		profile:               profile,
+		frontendUrl:           config.FrontendConfig.URL,
+		fhirURL:               fhirURL,
+		transport:             localFHIRStoreTransport,
+		workflows:             taskengine2.DefaultWorkflows(),
```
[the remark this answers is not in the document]
dwelman: Eww agreed, not sure why it did this but I will fix

orchestrator/careplancontributor/service.go
dwelman: Ah fair point! I will add some additional checks and tests

orchestrator/careplancontributor/service.go
dwelman: As discussed, I will add improved logging around message delivery in general

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -84,7 +94,7 @@ func (s *Service) markPrimaryTaskAsCompleted(subTask *fhir.Task) error {
 	}
 
 	var partOfTask fhir.Task
-	err = s.fhirClient.Read(*partOfTaskRef, &partOfTask)
+	err = s.carePlanServiceClient.Read(*partOfTaskRef, &partOfTask)
```
reinkrul, line 97: but it should only do `markPrimaryTaskAsCompleted` if it's the Task.owner?

reinkrul (commented):
I think there's also the check that it should only finish the primary task if it's the Task.owner?

reinkrul (approved):

PR #87, opened 2024-10-28
INT-385: Ensuring the SearchParameter used by the CPS via the code. 

JorisHeadease:
This is added when the CPS is enabled in the config. It performs via a conditional create. Removed the "init" scripts, and reconfigured to docker-compose file to add a health-check service for HAPI, so that docker services and also depend on a "live" FHIR server.


orchestrator/careplanservice/service.go
```diff
@@ -443,3 +445,51 @@ func (s Service) baseUrl() *url.URL {
...
 	return strings.Split(resourcePath, "/")[0]
 }
+
+func (s *Service) ensureSearchParameterExists() {
+	searchParamID := "CarePlan-subject-identifier"
+
+	// Define the SearchParameter
+	searchParam := fhir.SearchParameter{
+		Id:          &searchParamID,
+		Url:         "http://zorgbijjou.nl/SearchParameter/CarePlan-subject-identifier",
```
reinkrul, line 455: out of scope for this change, but we should probably change this to Shared Care Planning.

reinkrul (approved):

orchestrator/careplanservice/service.go
dwelman, line 455: I feel like we are limiting ourselves by hardcoding these endpoints and email addresses for future testing, environments, and scaling. Should this not be configurable?

JorisHeadease: We could, but in FHIR it's "just" a [canonical identifier](https://www.hl7.org/fhir/r4/references.html#canonical), so they may/should not change over time. It's also just created once, when the server starts up and it's not present yet.

dwelman: Ok, sounds good to me

PR #90, opened 2024-10-30
Fix questionnaire subtask filling

dwelman:
Dev testing was broken as we could create tasks but would not see the questionnaire, this was due to the task filler not being notified due to an incorrect endpoint and validation logic that would reject it.

Endpoint has been updated to the correct one, and new validation logic for SCP subtasks has been implemented

### DIFF
--- a/orchestrator/careplanservice/handle_createtask.go
+++ b/orchestrator/careplanservice/handle_createtask.go
@@ -25,6 +25,10 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 		return nil, fmt.Errorf("invalid %T: %w", task, err)
 	}
 
+	if !coolfhir.IsScpTask(&task) {
+		return nil, coolfhir.NewErrorWithCode("Task is not SCP task", http.StatusBadRequest)
+	}
+
 	switch task.Status {
 	case fhir.TaskStatusRequested:
 	case fhir.TaskStatusReady:
@@ -115,16 +119,16 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 				return nil, err
 			}
 			// Verify the owner of the parent task is the same as the org creating the subtask
-			isOwner, _ := coolfhir.ValidateTaskOwnerAndRequester(&parentTask, principal.Organization.Identifier)
+			isOwner, _ := coolfhir.IsIdentifierTaskOwnerAndRequester(&parentTask, principal.Organization.Identifier)
 			if !isOwner {
 				return nil, coolfhir.NewErrorWithCode("requester is not the owner of the parent task", http.StatusUnauthorized)
 			}
 
 			if !coolfhir.LogicalReferenceEquals(*parentTask.Owner, *task.Requester) {
-				return nil, coolfhir.NewErrorWithCode("requester is not the same as the parent task owner", http.StatusUnauthorized)
+				return nil, coolfhir.NewErrorWithCode("requester is not the same as the parent task owner", http.StatusBadRequest)
 			}
 			if !coolfhir.LogicalReferenceEquals(*parentTask.Requester, *task.Owner) {
-				return nil, coolfhir.NewErrorWithCode("owner is not the same as the parent task requester", http.StatusUnauthorized)
+				return nil, coolfhir.NewErrorWithCode("owner is not the same as the parent task requester", http.StatusBadRequest)
 			}
 		} else {
 			// we have a valid reference to a CarePlan, use this to retrieve the CarePlan and CareTeam to validate the requester is a participant
--- a/orchestrator/careplanservice/handle_createtask.go
+++ b/orchestrator/careplanservice/handle_createtask.go
@@ -96,19 +96,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 		if err != nil {
 			return nil, fmt.Errorf("invalid Task.basedOn: %w", err)
 		}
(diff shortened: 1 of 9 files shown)


orchestrator/careplanservice/handle_createtask.go
```diff
@@ -96,19 +96,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
+				return nil, err
+			}
+			// Verify the owner of the parent task is the same as the org creating the subtask
+			isOwner, _ := coolfhir.ValidateTaskOwnerAndRequester(&parentTask, principal.Organization.Identifier)
+			if !isOwner {
+				return nil, coolfhir.NewErrorWithCode("requester is not the owner of the parent task", http.StatusUnauthorized)
+			}
+
+			if !coolfhir.LogicalReferenceEquals(*parentTask.Owner, *task.Requester) {
+				return nil, coolfhir.NewErrorWithCode("requester is not the same as the parent task owner", http.StatusUnauthorized)
```
JorisHeadease, line 124: I would return a 422 or 400 here, they are authorized to do so, but they provided an incorrect Task

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -96,19 +96,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
+			}
+			if !coolfhir.LogicalReferenceEquals(*parentTask.Requester, *task.Owner) {
+				return nil, coolfhir.NewErrorWithCode("owner is not the same as the parent task requester", http.StatusUnauthorized)
```
JorisHeadease, line 127: Same on the status code

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -96,19 +96,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
 		}
-		if len(careTeams) == 0 {
-			return nil, coolfhir.NewErrorWithCode("CareTeam not found in bundle", http.StatusNotFound)
-		}
 
-		participant := coolfhir.FindMatchingParticipantInCareTeam(careTeams, principal.Organization.Identifier)
-		if participant == nil {
-			return nil, coolfhir.NewErrorWithCode("requester is not part of CareTeam", http.StatusUnauthorized)
+		// Different validation logic for an SCP subtask
+		if coolfhir.IsScpTask(&task) && len(task.PartOf) > 0 {
```
JorisHeadease, line 107: The `coolfhir.IsScpTask(&task)` part should also apply to the `else`-scope I would say. We probably want to throw an error when a non-SCP Task is sent to the CPS?

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -96,19 +96,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
+				return nil, coolfhir.NewErrorWithCode("SCP subtask must have exactly one parent task", http.StatusBadRequest)
+			}
+			// Get the parent task
+			var parentTask fhir.Task
+			err = s.fhirClient.Read(*task.PartOf[0].Reference, &parentTask)
+			if err != nil {
+				return nil, err
+			}
+			// Verify the owner of the parent task is the same as the org creating the subtask
+			isOwner, _ := coolfhir.ValidateTaskOwnerAndRequester(&parentTask, principal.Organization.Identifier)
```
JorisHeadease, line 118: Maybe out-of-scope for this PR, but this function name confused me. It doesn't "validate" anything right?

JorisHeadease (commented):
Looks good, added some small comments

orchestrator/careplanservice/handle_createtask.go
dwelman: I guess not. It is used * for * validation, in that it confirms if a user is owner/requester. If it is causing confusion then I definitely think it can be renamed.

Maybe something like `IsIdentifierTaskOwnerAndRequester`?

orchestrator/careplanservice/handle_createtask.go
dwelman: Sounds good. I initially thpought we should support it since Task Filler skips it which made me think it was possible to create a non-SCP task but have it just not go through the task filler logic. I will add a check earlier in the code

JorisHeadease: Yeah not a 100% sure, but it feels weird sending a non-SCP Task to the CPS. Maybe best to be restrictive for now, and we'll find out soon enough if it is required 😂

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -96,19 +100,46 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
 		}
-		if len(careTeams) == 0 {
-			return nil, coolfhir.NewErrorWithCode("CareTeam not found in bundle", http.StatusNotFound)
-		}
 
-		participant := coolfhir.FindMatchingParticipantInCareTeam(careTeams, principal.Organization.Identifier)
-		if participant == nil {
-			return nil, coolfhir.NewErrorWithCode("requester is not part of CareTeam", http.StatusUnauthorized)
+		// Different validation logic for an SCP subtask
+		if coolfhir.IsScpTask(&task) && len(task.PartOf) > 0 {
```
JorisHeadease, line 111:
```suggestion
		if len(task.PartOf) > 0 {
```

JorisHeadease (approved):

PR #92, opened 2024-11-01
Implement task and subtask updating and resolution

dwelman:
There was an issue where the dev test would hang because the primary task would not update and the site would be stuck in a loop requesting the task and waiting for it to be marked completed

There is still a strange bug as it tries to load a questionnaire screen that no longer exists, might need to be fixed on FE but confirming


dwelman:
Accidentally included some debug logging, these will be removed before merging

dwelman:
I have temporarily disabled the state transition check as that is also causing issues with the flow, we need to determine how to make sure CPC is sending the right state to CPS

reinkrul:
Removed internal link

PR #96, opened 2024-11-01
[INT-398] If bundle in CarePlan search is empty, return early

dwelman:
We were searching for a CareTeam in an empty bundle, causing failures

### DIFF
--- a/deployments/dev/start.sh
+++ b/deployments/dev/start.sh
@@ -141,7 +141,6 @@ issueUraCredential "clinic" "${CLINIC_DID}" "1234" "Demo Clinic" "Utrecht"
 echo "  Registering on Nuts Discovery Service"
 curl -X POST -H "Content-Type: application/json" -d "{\"registrationParameters\":{\"fhirNotificationURL\": \"${CLINIC_URL}/fhir/notify\"}}" http://localhost:8081/internal/discovery/v1/dev:HomeMonitoring2024/clinic
 echo "  Waiting for the FHIR server to be ready"
-./config/init-fhir-resources.sh $HOSPITAL_URL
 popd
 
 echo "Creating stack for Clinic..."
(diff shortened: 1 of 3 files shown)


deployments/dev/start.sh
```diff
@@ -141,7 +141,6 @@ issueUraCredential "clinic" "${CLINIC_DID}" "1234" "Demo Clinic" "Utrecht"
 echo "  Registering on Nuts Discovery Service"
 curl -X POST -H "Content-Type: application/json" -d "{\"registrationParameters\":{\"fhirNotificationURL\": \"${CLINIC_URL}/fhir/notify\"}}" http://localhost:8081/internal/discovery/v1/dev:HomeMonitoring2024/clinic
 echo "  Waiting for the FHIR server to be ready"
-./config/init-fhir-resources.sh $HOSPITAL_URL
```
dwelman, line 144: This was removed in an earlier commit and snuck back in during a merge conflict, removing again so we can run the local test stack

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -47,6 +48,12 @@ func (s *Service) handleSearchCarePlan(ctx context.Context, queryParams url.Valu
 	if err != nil {
 		return nil, err
 	}
+
+	if len(bundle.Entry) == 0 {
+		log.Info().Msg("CarePlan search returned empty bundle")
```
reinkrul, line 53: Would this occur commonly? I think not something you want to log on INFO (if you want to log it at all)

reinkrul (approved):

orchestrator/careplanservice/handle_getcareplan.go
dwelman: Hmm that's a fair point. I picked it up happening because a search for CarePlan would be done before the Task was created (I think for the CarePlan selection)

PR #97, opened 2024-11-01
INT-395: Set right content-type for bundle responses, deduplicate response marshalling

reinkrul:
Also removed some unused code.

### DIFF
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -251,16 +251,11 @@ func (s *Service) handleCreateOrUpdate(httpRequest *http.Request, httpResponse h
 		log.Logger.Warn().Msgf("Failed to parse status code from transaction result (responding with 200 OK): %s", fhirResponse.Status)
 		statusCode = http.StatusOK
 	}
-	httpResponse.Header().Add("Content-Type", coolfhir.FHIRContentType)
-	if fhirResponse.Location != nil {
-		httpResponse.Header().Add("Location", *fhirResponse.Location)
-	}
-	// TODO: I won't pretend I tested the other response headers (e.g. Last-Modified or ETag), so we won't set them for now.
-	//       Add them (and test them) when needed.
-	httpResponse.WriteHeader(statusCode)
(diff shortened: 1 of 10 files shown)


orchestrator/careplanservice/service.go
```diff
@@ -284,18 +279,11 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
 		coolfhir.WriteOperationOutcomeFromError(err, operationName, httpResponse)
 		return
 	}
-
+	hdrs := make(map[string]string)
```
dwelman, line 282:
```suggestion
	outputHeaders := make(map[string]string)
```
I know we already have `headers` but I initially wasn't sure what `hdrs` were

orchestrator/careplanservice/service.go
```diff
@@ -284,18 +279,11 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
...
 	for key, value := range headers.Header {
-		httpResponse.Header()[key] = value
-	}
-
-	b, err := json.Marshal(resource)
-	_, err = httpResponse.Write(b)
-	if err != nil {
-		coolfhir.WriteOperationOutcomeFromError(err, operationName, httpResponse)
-		return
+		hdrs[key] = value[0]
```
dwelman, line 284:
```suggestion
		outputHeaders[key] = value[0]
```

orchestrator/careplanservice/service.go
```diff
@@ -284,18 +279,11 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
...
 	}
-	return
+	coolfhir.SendResponse(httpResponse, http.StatusOK, resource, hdrs)
```
dwelman, line 286:
```suggestion
	coolfhir.SendResponse(httpResponse, http.StatusOK, resource, outputHeaders)
```

dwelman (approved):

dwelman:
Really nice refactor! Glad you could remove so much code and replace it with a much simpler method

PR #98, opened 2024-11-02
INT-316: Use correct fields for selecting questionnaires

reinkrul:
Changes the following:
- Fixes selection of the questionnaires to correctly use `Task.reasonCode` (code of condition) and `Task.focus(serviceRequest).code` (telemonitoring).
- Improves Task Engine unit tests (more homogenous, more tests)
- Only accept primary task if it's in state `received` or `requested`.


orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -37,9 +37,9 @@ func (s *Service) handleTaskFillerCreateOrUpdate(ctx context.Context, cpsClient
 	// If partOfRef is nil, handle the task as a primary task - no need to create follow-up subtasks for newly created Tasks
 	//This only happens on Task update where the Task.output is filled with a QuestionnaireResponse
 	if partOfRef == nil {
-		// Check if the priary task is "created", its status will be updated by subtasks that are completed - not directly here
-		if task.Status != fhir.TaskStatusRequested {
-			log.Debug().Msg("primary Task.status != requested (workflow already started) - not processing in handleTaskFillerCreateOrUpdate")
+		// Check if the primary task is "created", its status will be updated by subtasks that are completed - not directly here
+		if task.Status != fhir.TaskStatusRequested && task.Status != fhir.TaskStatusReceived {
```
dwelman, line 41: Considering we use this again lower down, do we maybe want to centralise this condition somewhere? Will make future updates easier as well if we want to exclude more statuses

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -258,6 +265,77 @@ func (s *Service) createSubTaskOrFinishPrimaryTask(cpsClient fhirclient.Client,
...
 }
 
+// selectWorkflow determines the workflow to use based on the Task's focus, and reasonCode or reasonReference.
+// It first selects the type of service, from the Task.focus (ServiceRequest), and then selects the workflow based on the Task.reasonCode or Task.reasonReference.
+// If it finds no, or multiple, matching workflows, it returns an error.
+func (s *Service) selectWorkflow(cpsClient fhirclient.Client, task *fhir.Task) (*taskengine.Workflow, error) {
+	var matchedServiceCodes []string
+	// Determine service code from Task.focus
+	var serviceRequest fhir.ServiceRequest
+	if err := cpsClient.Read(*task.Focus.Reference, &serviceRequest); err != nil {
```
dwelman, line 275: We do not have an explicit endpoint for GET ServiceRequest, won't this fail in test/prod with the most recent change to deny unmanaged operations?

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -258,6 +265,77 @@ func (s *Service) createSubTaskOrFinishPrimaryTask(cpsClient fhirclient.Client,
...
+		for _, coding := range task.ReasonCode.Coding {
+			if coding.System == nil || coding.Code == nil {
+				continue
+			}
+			candidateCodes = append(candidateCodes, coding)
+		}
+	}
+	if task.ReasonReference != nil && task.ReasonReference.Reference != nil {
+		var condition fhir.Condition
+		if err := cpsClient.Read(*task.ReasonReference.Reference, &condition); err != nil {
```
dwelman, line 306: Same question as above with ServiceRequest

orchestrator/careplancontributor/handle_taskfiller.go
JorisHeadease: Doesn't this create a subtask twice when the primary Task goes from `requested` to `received`? Creating a new Task with the status `received` is not allowed:

JorisHeadease (commented):
Started the review, will continue tomorrow

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -260,6 +267,77 @@ func (s *Service) createSubTaskOrFinishPrimaryTask(cpsClient fhirclient.Client,
...
+	if task.ReasonCode != nil {
+		for _, coding := range task.ReasonCode.Coding {
+			if coding.System == nil || coding.Code == nil {
+				continue
+			}
+			candidateCodes = append(candidateCodes, coding)
+		}
+	}
+	if task.ReasonReference != nil && task.ReasonReference.Reference != nil {
+		var condition fhir.Condition
```
reinkrul, line 307: if it's not a Condition, ignore it

orchestrator/careplancontributor/handle_taskfiller_test.go
```diff
@@ -3,171 +3,257 @@ package careplancontributor
...
+						},
+						{
+							System: to.Ptr("http://snomed.info/sct"),
+							Code:   to.Ptr("84114007"), // Heart failure
+						},
+					},
+				}
+			}),
+			expectedError: errors.New("failed to process new primary Task: Task.reasonCode or Task.reasonReference matches multiple workflows (http://snomed.info/sct|13645005, http://snomed.info/sct|84114007) for service http://snomed.info/sct|719858009"),
+		},
```
JorisHeadease, line 94: Might be good to also add a case that proves duplicate codes (two COPD) are fine, as long as it results in one reasonCodes

orchestrator/careplancontributor/handle_taskfiller_test.go
```diff
@@ -3,171 +3,257 @@ package careplancontributor
...
+			notificationTask: deep.AlterCopy(primaryTask, func(t *fhir.Task) {
+				t.Status = fhir.TaskStatusInProgress
+			}),
+		},
+		{
+			name: "error: primary task, unknown service is requested (primary Task.focus(ServiceRequest).code is not supported)",
+			serviceRequest: deep.AlterCopy(serviceRequest, func(sr *fhir.ServiceRequest) {
+				sr.Code.Coding[0].Code = to.Ptr("UnknownServiceCode")
+			}),
+			expectedError: errors.New("failed to process new primary Task: ServiceRequest.code does not match any offered services"),
```
JorisHeadease, line 174: I think we're missing a test where the same check is done for the "Task.reasonCode or Task.reasonReference does not match any workflow for service"

JorisHeadease (commented):
nice PR @reinkrul! I've added a commit to make sure the Zorgplatform ServiceRequest matches the Telemonitoring service. As well, added 2 comments for unit tests that can be added still. LGTM though!

orchestrator/careplancontributor/handle_taskfiller.go
reinkrul: Yes, and ServiceRequest (and Condition) are (or should be) on the list to be fixed/authorized. Test environment currently still enables unmanaged FHIR operations.

orchestrator/careplancontributor/handle_taskfiller.go
reinkrul: > Considering we use this again lower down, do we maybe want to centralise this condition somewhere? Will make future updates easier as well if we want to exclude more statuses

Yes, and I think the task engine can use some cleanup/refactoring, as it has become quite the complex beast. But let's do that in a different PR.

dwelman (approved):

PR #103, opened 2024-11-06
[INT-370] Create logout endpoint

dwelman:
Add endpoint for logout which destroys the current session and redirects the user back to the base frontend URL

Still TODO is to redirect to a real logout page, which has not been created yet

### DIFF
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -134,7 +134,7 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 		s.SessionManager.Destroy(writer, request)
 		// If there is a 'Referrer' value in the header, redirect to that URL
 		if referrer := request.Header.Get("Referrer"); referrer != "" {
-			http.Redirect(writer, request, referrer, http.StatusOK)
+			http.Redirect(writer, request, referrer, http.Found)
 		} else {
 			// This redirection will be handled by middleware in the frontend
 			http.Redirect(writer, request, s.frontendUrl, http.StatusOK)
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -132,9 +132,9 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	// Logout endpoint
 	mux.HandleFunc(basePath+"/logout", s.withSession(func(writer http.ResponseWriter, request *http.Request, session *user.SessionData) {
 		s.SessionManager.Destroy(writer, request)
-		// If there is a 'Referrer' value in the header, redirect to that URL
-		if referrer := request.Header.Get("Referrer"); referrer != "" {
-			http.Redirect(writer, request, referrer, http.Found)
+		// If there is a 'Referer' value in the header, redirect to that URL
+		if referer := request.Header.Get("Referer"); referer != "" {
+			http.Redirect(writer, request, referer, http.StatusFound)
 		} else {
 			// This redirection will be handled by middleware in the frontend
 			http.Redirect(writer, request, s.frontendUrl, http.StatusOK)
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -130,10 +130,16 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 	})
 
 	// Logout endpoint
-	mux.HandleFunc(basePath+"/zorgplatform/logout", func(response http.ResponseWriter, request *http.Request) {
-		s.SessionManager.Destroy(response, request)
-		http.Redirect(response, request, s.frontendUrl, http.StatusOK)
-	})
+	mux.HandleFunc(basePath+"/zorgplatform/logout", s.withSession(func(writer http.ResponseWriter, request *http.Request, session *user.SessionData) {
+		s.SessionManager.Destroy(writer, request)
+		// If there is a 'Referrer' value in the header, redirect to that URL
+		if referrer := request.Header.Get("Referrer"); referrer != "" {
+			http.Redirect(writer, request, referrer, http.StatusOK)
+		} else {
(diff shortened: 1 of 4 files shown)


orchestrator/careplancontributor/service.go
```diff
@@ -128,6 +128,12 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 		log.Info().Msgf("Redirecting to %s", s.frontendUrl)
 		http.Redirect(response, request, s.frontendUrl, http.StatusFound)
 	})
+
+	// Logout endpoint
+	mux.HandleFunc(basePath+"/zorgplatform/logout", func(response http.ResponseWriter, request *http.Request) {
+		s.SessionManager.Destroy(response, request)
+		http.Redirect(response, request, s.frontendUrl, http.StatusOK)
+	})
```
JorisHeadease, line 136: I see the `zorgplatform` is defined in the issue, but it feels like we either:

1. Want a generic logout, no matter how you launched orca, OR
2. this logic should be moved into the zorgplatform `service.go`

I'd prefer option 1, maybe the `zorgplatform` can be replaced with a wildcard?

orchestrator/user/session.go
```diff
@@ -62,6 +62,20 @@ func (m *SessionManager) Get(request *http.Request) *SessionData {
 	return &session.Data
 }
 
+func (m *SessionManager) Destroy(response http.ResponseWriter, request *http.Request) {
+	sessionID := getSessionCookie(request)
+	if sessionID != "" {
+		m.store.destroy(sessionID)
+	}
```
JorisHeadease, line 69: We should probably do something when the session is not found, either log a warning or maybe even return an error?

orchestrator/careplancontributor/service.go
```diff
@@ -128,6 +128,12 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 		log.Info().Msgf("Redirecting to %s", s.frontendUrl)
 		http.Redirect(response, request, s.frontendUrl, http.StatusFound)
 	})
+
+	// Logout endpoint
+	mux.HandleFunc(basePath+"/zorgplatform/logout", func(response http.ResponseWriter, request *http.Request) {
```
reinkrul, line 133: I'd have this as authenticated endpoint (`s.withSession`), then you know for sure there's actually a session to destroy

orchestrator/user/session.go
```diff
@@ -62,6 +62,20 @@ func (m *SessionManager) Get(request *http.Request) *SessionData {
 	return &session.Data
 }
 
+func (m *SessionManager) Destroy(response http.ResponseWriter, request *http.Request) {
```
reinkrul, line 65: needs unit test

reinkrul (changes requested):
see comments

orchestrator/careplancontributor/service.go
reinkrul: I'm not 100% convinced, since the IDP sends (or should sent, according to SAMLv2 spec) a signed `LogoutRequest`, which' signature we _could_. I don't think it's required, but maybe in future?

orchestrator/careplancontributor/service.go
```diff
@@ -128,6 +128,12 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
 		log.Info().Msgf("Redirecting to %s", s.frontendUrl)
 		http.Redirect(response, request, s.frontendUrl, http.StatusFound)
```
reinkrul, line 129: Don't redirect to the frontend URL, because the it shows the frontend enrolment UI without any data (or even "no user session", or something). Either;
- Redirect to URL in `Referer` header, or
- Render some text that says "you have been logged out" if there's no Referer header.

orchestrator/user/session.go
dwelman: For now let's log a warning, we can use those logs to determine how often it is happening and if we need to error out.

For now I don't see any harm in logout being called without a session open but maybe I am missing something?

orchestrator/careplancontributor/service.go
dwelman: This would be future implementation right?

reinkrul: I think we can do with 1 logout URL (having it on `/logout`?), since we don't need to check the signature right now. And session destroying isn't adapter-specific (especially if we can just redirect back to the Referer URL).

orchestrator/careplancontributor/service.go
```diff
@@ -128,6 +128,18 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
...
 	})
+
+	// Logout endpoint
+	mux.HandleFunc(basePath+"/logout", s.withSession(func(writer http.ResponseWriter, request *http.Request, session *user.SessionData) {
+		s.SessionManager.Destroy(writer, request)
+		// If there is a 'Referrer' value in the header, redirect to that URL
+		if referrer := request.Header.Get("Referrer"); referrer != "" {
```
reinkrul, line 136: it's `Referer` (USA/UK English?): https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

reinkrul (changes requested):

orchestrator/careplancontributor/service.go
```diff
@@ -128,6 +128,18 @@ func (s Service) RegisterHandlers(mux *http.ServeMux) {
...
+			http.Redirect(writer, request, referrer, http.StatusOK)
```
reinkrul, line 137:
```suggestion
			http.Redirect(writer, request, referrer, http.Found)
```

orchestrator/careplancontributor/service.go
dwelman: Ahhh. That is US english, I have learnt how to spell with UK/SA english :P will fix. 'Color' also always gets me...

reinkrul (approved):

PR #104, opened 2024-11-06
Frontend: task performer as display name and/or URA

reinkrul:
If no display name, then at least it shows the URA number as "URA: 1234". Or it shows the full coding system value if it's not a URA.


reinkrul:
Oh thanks @JorisHeadease 😆

JorisHeadease (approved):
Changed the OrganizationLabel into a React component.

PR #107, opened 2024-11-07
[INT-405] Disallow unmanaged FHIR operations in CPS

dwelman:
Add a flag that allows unmanaged FHIR operations, default behaviour is that they are not allowed

### DIFF
--- a/orchestrator/careplanservice/config.go
+++ b/orchestrator/careplanservice/config.go
@@ -13,7 +13,7 @@ type Config struct {
 	FHIR    coolfhir.ClientConfig `koanf:"fhir"`
 	Enabled bool                  `koanf:"enabled"`
 	// Defaults to false, should not be set true in Test or Prod
-	AllowUnmanagedFHIROperations bool `koanf:"allow_unmanaged_fhir_operations"`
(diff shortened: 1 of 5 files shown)


orchestrator/careplanservice/config.go
```diff
@@ -12,6 +12,8 @@ func DefaultConfig() Config {
 type Config struct {
 	FHIR    coolfhir.ClientConfig `koanf:"fhir"`
 	Enabled bool                  `koanf:"enabled"`
+	// Defaults to false, should not be set true in Test or Prod
+	AllowUnmanagedFHIROperations bool `koanf:"allow_unmanaged_fhir_operations"`
```
reinkrul, line 16: underscores break koanf, since it'll think it's a sub-struct (meaning nested Allow, Unmanaged, Fhir, and Operations structs), since it uses that delimiter for nesting when specifying them as environment variables.

reinkrul (changes requested):

orchestrator/careplanservice/config.go
dwelman: Good to know, thanks. Will update and fix the e2e test

reinkrul (approved):

PR #108, opened 2024-11-07
Validating Zorgplatform signatures 

JorisHeadease:
Validating Zorgplatform signatures that are:

- sent during the SSO launch 
- while receiving a RequestSecurityTokenResponse (requesting an access_token)


dwelman (approved):
LGTM though I also don't have too much experience working with certs so I can't say for certain

orchestrator/careplancontributor/applaunch/zorgplatform/handle_create_hcp_rst_test.go
```diff
@@ -192,10 +195,24 @@ func TestService_RequestHcpRst_IntegrationTest(t *testing.T) {
 		t.Skip("Skipping TestService_RequestHcpRst_IntegrationTest as test-sign-zorgplatform.private.pem is not present locally")
 	}
 
+	zorgplatformCertData, err := os.ReadFile("zorgplatform.online.pem")
+	if err != nil {
```
reinkrul, line 199: use `require.NoError()`

reinkrul (changes requested):
This part is really sensitive (bug would allow attacker to just log into ORCA and query CPS, other CPCs and local EHR), so I expect test(s) that test bad/crafted (signed with other private key) as well.

reinkrul (approved):

PR #110, opened 2024-11-07
[INT-410] Prune user sessions every minute to delete expired sessions

dwelman:


orchestrator/cmd/server.go
```diff
@@ -59,6 +59,21 @@ func Start(config Config) error {
 			}
 			services = append(services, service)
 		}
+
+		// Start session expiration ticker
+		ticker := time.NewTicker(time.Minute)
+		shutdown := make(chan struct{})
```
reinkrul, line 65: nothing is ever actually sent to this channel?

In theory, we should do cleanup with `Stop()`, but there will only ever be 1 instance of this Ticker which lives until the application shuts down, so I don't really worry about properly stopping it..?

reinkrul (changes requested):

orchestrator/cmd/server.go
dwelman: Ah that is a good point, forgot to actually use the shutdown channel... Yeah, seems fine to remove it then

reinkrul: Typically you'd capture kill signal, then send something down that channel. But we don't need to clean anything up, so not necesarry.

reinkrul (approved):

PR #115, opened 2024-11-07
[INT-405] Deny unmanaged GET operations

dwelman:


orchestrator/careplanservice/service.go
```diff
@@ -284,6 +284,13 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
 		resource, err = s.handleGetTask(httpRequest.Context(), resourceId, headers)
 	default:
 		log.Warn().Msgf("Unmanaged FHIR operation at CarePlanService: %s %s", httpRequest.Method, httpRequest.URL.String())
+		if !s.allowUnmanagedFHIROperations {
```
reinkrul, line 287: we ahve this at 3 places now I think? Maybe we should centralize this functionality, since it's becoming risky to have the same code (important for security) at 3 places?

dwelman: Yes that's reasonable, especially because this is security related code

dwelman: I looked at where to centralise but it is not so straight forward as we have different allowed operations for GET (CP, CT, Task) and PUT (Task) so we can't have one higher level check that validates type + allowed operations without adding duplicate logic checks that we already have in place for HTTP op + Resource type. 

I have at least centralised checking the boolean, is that alright or should we further centralise?

reinkrul (approved):

PR #116, opened 2024-11-07
Update query parameter handling in care services

rolandgroen:
Refactor query parameter loops to handle multiple values per key in the `handleSearchTask`, `handleSearchCareTeam`, and `handleSearchCarePlan` functions. This ensures that all values for a given query parameter key are included, improving the handling of multi-value parameters.

### DIFF
--- a/orchestrator/careplanservice/handle_getcareplan.go
+++ b/orchestrator/careplanservice/handle_getcareplan.go
@@ -39,8 +39,8 @@ func (s *Service) handleSearchCarePlan(ctx context.Context, queryParams url.Valu
 			if k == "_include" && value == "CarePlan:care-team" {
 				continue
 			}
+			params = append(params, fhirclient.QueryParam(k, value))
 		}
-		params = append(params, fhirclient.QueryParam(k, v[0]))
 	}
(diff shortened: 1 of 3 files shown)


orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -34,9 +34,11 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
-		// Skip param to include CareTeam since we need to add this for validation anyway
-		if k == "_include" && v[0] == "CarePlan:care-team" {
-			continue
+		for _, value := range v {
+			// Skip param to include CareTeam since we need to add this for validation anyway
+			if k == "_include" && value == "CarePlan:care-team" {
+				continue
+			}
 		}
 		params = append(params, fhirclient.QueryParam(k, v[0]))
```
dwelman, line 43: This should be removed and `params = append(params, fhirclient.QueryParam(k, value))` should be added into the range loop otherwise we might run into this problem again in the future with CarePlan search parameters

dwelman (approved):

PR #119, opened 2024-11-08
[INT-411] Disable CPC zorginzage endpoint by default

dwelman:
A flag can be provided to enable this


orchestrator/careplancontributor/config.go
```diff
@@ -14,12 +14,13 @@ func DefaultConfig() Config {
...
-	AppLaunch         applaunch.Config      `koanf:"applaunch"`
-	FHIR              coolfhir.ClientConfig `koanf:"fhir"`
-	Enabled           bool                  `koanf:"enabled"`
-	StaticBearerToken string
+	CarePlanService           CarePlanServiceConfig `koanf:"careplanservice"`
+	FrontendConfig            FrontendConfig        `koanf:"frontend"`
+	AppLaunch                 applaunch.Config      `koanf:"applaunch"`
+	FHIR                      coolfhir.ClientConfig `koanf:"fhir"`
+	Enabled                   bool                  `koanf:"enabled"`
+	ZorginzageEndpointEnabled bool                  `koanf:"zorginzageendpointenabled"`
```
reinkrul, line 22: can you add a unit test for `DefaultConfig()` that verifies it's disabled by default? Since it's important for security.

reinkrul (changes requested):
I think we shouldn't call it "zorginzage", as that's Dutch and a specific use case. Maybe you can come up with a more generic, internationalized name for it?

orchestrator/careplancontributor/config.go
dwelman: Good callout, will do alongside the change to use an English term

orchestrator/careplancontributor/config_test.go
```diff
@@ -17,3 +17,10 @@ func TestConfig_Validate(t *testing.T) {
 		require.EqualError(t, err, "careplancontributor.careplanservice.url is not configured")
 	})
 }
+
+func TestDefaultConfig(t *testing.T) {
+	t.Run("validate healthDataView Endpoint is disabled by default", func(t *testing.T) {
+		config := DefaultConfig()
+		require.Equal(t, false, config.HealthDataViewEndpointEnabled)
```
reinkrul, line 24: assert.False (or require), but really trivial

reinkrul (approved):

PR #121, opened 2024-11-08
INT-391: check ID of task being updated to prevent accidentally/intentionally updating other Tasks

reinkrul:
I believe compliant FHIR APIs should prevent this (I believe I saw an error of HAPI once), but to be 100% sure we can just check it. Otherwise, could be misuse to secretly update another task that the caller isn't authorized to.

@bramwesselo or should be leave this up to HAPI/Azure FHIR?


dwelman (approved):

JorisHeadease:
Ah, just got merged during my review I see @reinkrul

PR #124, opened 2024-11-11
Added logic to pre-select a Condition by setting the ServiceRequest.reasonReference

JorisHeadease:
Also added Zorgplatform "magic" fetch the Condition based on this reasonReference, which can be cached. The "fractuur van pols" has been replaced with "Hartfalen" for demo purposes.

### DIFF
--- a/frontend/app/enrollment/components/enrollment-details-view.tsx
+++ b/frontend/app/enrollment/components/enrollment-details-view.tsx
@@ -40,7 +40,8 @@ export default function EnrollmentDetailsView() {
                     <div className='font-bold'>Service:</div>
                     <div className='col-span-3'>{serviceRequest?.code?.coding?.[0].display || "Unknown"}</div>
                     <div className='font-bold'>Care Path:</div>
-                    <div className='col-span-3'>{taskCondition?.code?.coding?.[0].display || "Unknown"}</div>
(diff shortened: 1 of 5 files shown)


frontend/app/enrollment/components/enrollment-details-view.tsx
```diff
@@ -40,7 +40,8 @@ export default function EnrollmentDetailsView() {
                     <div className='font-bold'>Service:</div>
                     <div className='col-span-3'>{serviceRequest?.code?.coding?.[0].display || "Unknown"}</div>
                     <div className='font-bold'>Care Path:</div>
-                    <div className='col-span-3'>{taskCondition?.code?.coding?.[0].display || "Unknown"}</div>
+                    {/* TODO: Check whether to prefer code.text or code.coding.display - NullFlavour hurts in coding display */}
```
reinkrul, line 43: both?

reinkrul (approved):

frontend/app/enrollment/components/enrollment-details-view.tsx
JorisHeadease: Yeah, can do. But it's a but ugly as they do contain values, like `other` that need to be checked/excluded

PR #128, opened 2024-11-12
Zpf bgz viewer

JorisHeadease:
Added a `/viewer/bgz` endpoint in the orca code that renders Zorgplatforms' OAS doc, but changed the following:

1. Points to the CPC EHR Proxy
2. Added a `POST` operation for the `Flag` resource
3. Removed the Authorizations, as we simply use the session to proxy data

Some requests do not work:

1. "FHIR operation 'Search' not supported for resource type 'MedicationDispense' and scope 'Bgz'."
2. The `/fhir/Observation/{id}` doesn't work, even when copy-pasting an `id` from the GET `/fhir/Observation`. 
3. POSTing a new `Flag`: "Saml authentication did not succeed."

However, I see the same behaviour when directly performing these requests on the Zorgplatform Developer Portal. So I assume I am doing something wrong, or there is a bug/wrong data.

https://github.com/user-attachments/assets/5ae993d9-f0bb-4264-a419-030b30249f33


reinkrul (approved):

JorisHeadease:
Merging for now, the POST Task still needs to be "fixed" in a sense that it cannot be POSTed with a workflow HCP token, but seems to require a more general service token.

PR #129, opened 2024-11-12
Orchestrator: cache CSD lookups

reinkrul:
Should speedup the enrolment process a bit. 30 seconds cache should be enough to long enough for most interactions.


dwelman (approved):
Nice optimisation!

PR #133, opened 2024-11-13
[FIX] Adds result filtering to search for CarePlan, CareTeam, Task

dwelman:
Found an exploit as it was possible to access resources the requester should not have access to using the query params of a search.

This is a first pass on tightening auth, and I would recommend further tightening by limiting what we return in a search

NB: Still need to add Task filtering, putting the review out early for comments


reinkrul (commented):
Is there a way we can make it less copy/paste? More like a pipeline/middleware-kind of function? (filtering unwanted and unauthorized resources)?

dwelman:
@reinkrul yes that should be possible, I was hoping to start off simple with this and refactor while implementing the other resource types to see what patterns and similarities emerge. Would that be fine?

dwelman:
@reinkrul I have reduced some of the code duplication by making the initial bundle search generic.

Looking into making the filtering generic now, a bit more tricky since each resource has a different filter requirement but I am sure it can be done to some degree

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -32,41 +41,58 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+	includeCareTeamInResponse := false
+	if slices.Contains(queryParams["_include"], "CarePlan:care-team") {
+		includeCareTeamInResponse = true
+	} else {
+		queryParams.Add("_include", "CarePlan:care-team")
+	}
 
-	if len(bundle.Entry) == 0 {
-		return &bundle, nil
+	queryParams.Add("_include", "CarePlan:care-team")
```
reinkrul, line 57: it's double now?

orchestrator/careplanservice/util.go
```diff
@@ -49,12 +64,32 @@ func (s *Service) getCarePlanAndCareTeams(carePlanReference string) (fhir.CarePl
...
+	}
+
+	params = append(params, fhirclient.ResponseHeaders(headers))
+	var bundle fhir.Bundle
+	err := s.fhirClient.Read(resourceType, &bundle, params...)
+	if err != nil {
+		return nil, &fhir.Bundle{}, err
+	}
+
+	resourceSlice := reflect.New(reflect.SliceOf(reflect.TypeOf((*T)(nil)).Elem())).Interface()
```
reinkrul, line 84: wow

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -32,41 +41,58 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
 // if the requester is a participant of one of the returned CareTeams, return the whole bundle, else error
 // Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
 func (s *Service) handleSearchCarePlan(ctx context.Context, queryParams url.Values, headers *fhirclient.Headers) (*fhir.Bundle, error) {
```
reinkrul, line 43: ouch, this just lists all CarePlans in the CPS? Isn't that going to be terribly slow? I can't think of another approach though... Maybe a custom SearchParameter in future.

orchestrator/careplanservice/util.go
```diff
@@ -77,3 +112,36 @@ func matchResourceIDs(request *FHIRHandlerRequest, idFromResource *string) error
...
+	newBundle := fhir.Bundle{
+		Entry: []fhir.BundleEntry{},
+	}
+
+	for i, entry := range bundle.Entry {
+		var resourceInBundle coolfhir.Resource
+		err := json.Unmarshal(entry.Resource, &resourceInBundle)
+		if err != nil {
+			// We don't want to fail the whole operation if one resource fails to unmarshal
+			log.Error().Msgf("filterMatchingResourcesInBundle: Failed to unmarshal resource: %v", err)
```
reinkrul, line 127: I think you could even add an OperationOutcome entry to the bundle, alternatively

orchestrator/careplanservice/handle_getcareplan.go
dwelman: Ah whoops, good catch

orchestrator/careplanservice/handle_getcareplan.go
dwelman: Well, yes if the user does not provide any params then it will search everything. Ideally I would hope they provide an ID or some other filter.

I suppose we could enforce that the user provides at least something that will limit the results but that sounds messy too

orchestrator/careplanservice/util.go
dwelman: I have been staring at reflection code for the past few days so this seemed completely fine to me but yes taking a step back trying to read this is a punch in the face xD I have simplified

PR #138, opened 2024-11-15
[INT-402] Add GET Condition, Patient, Questionnaire, QuestionnaireResponse, ServiceRequest

dwelman:
Refactor some shared auth code, heavily refactor Get unit tests to reuse shared testing infrastructure


orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	var condition fhir.Condition
+	err = s.fhirClient.Read("Condition/"+id, &condition, fhirclient.ResponseHeaders(headers))
+	if err != nil {
+		return nil, err
+	}
+
+	// if the condition is for a patient, fetch the patient. If the requester has access to the patient they also have access to the condition
+	if condition.Subject.Reference != nil && strings.HasPrefix(*condition.Subject.Reference, "Patient/") {
```
dwelman, line 27: This is incorrect, refactoring to work with BSN.

orchestrator/careplanservice/handle_get_test.go
```diff
@@ -0,0 +1,727 @@
...
+	}
+	for _, tt := range serviceRequestTests {
+		testHelperHandleGetResource[fhir.ServiceRequest](t, tt)
+	}
+	for _, tt := range conditionTests {
+		testHelperHandleGetResource[fhir.Condition](t, tt)
+	}
+}
+
+func testHelperHandleGetResource[T any](t *testing.T, params TestStruct[T]) {
```
reinkrul, line 628: I think this could be simplified by turning it inside-out: just make actual Test_XX (e.,g. Task, Patient, etc) functions (or tests within the test with `t.Run()`), and having this setup functionality called from each respective test. Since the logic for testing a particular resource is now spread over 2 spots (test case definition and handler calling in this function).

orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+		if len(bundle.Entry) == 0 {
+			return nil, &coolfhir.ErrorWithCode{
+				Message:    "User does not have access to Condition",
+				StatusCode: http.StatusUnauthorized,
+			}
+		}
+	} else {
+		return nil, &coolfhir.ErrorWithCode{
+			Message:    "Condition does not have Patient as subject, can't verify access",
```
reinkrul, line 39: I think we might be leaking security details here. I'd have 1 message returned here (the one from line 33), and log the inner reason.

orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -0,0 +1,45 @@
...
+	// if the condition is for a patient, fetch the patient. If the requester has access to the patient they also have access to the condition
+	if condition.Subject.Identifier != nil && *condition.Subject.Identifier.System == "http://fhir.nl/fhir/NamingSystem/bsn" {
+		bundle, err := s.handleSearchPatient(ctx, map[string][]string{"identifier": {patientBSNFromIdentifier(condition.Subject.Identifier)}}, headers)
+		if err != nil {
+			return nil, err
+		}
+		if len(bundle.Entry) == 0 {
+			return nil, &coolfhir.ErrorWithCode{
+				Message:    "User does not have access to Condition",
+				StatusCode: http.StatusUnauthorized,
```
reinkrul, line 34: shouldn't this be Forbidden?

orchestrator/careplanservice/handle_getpatient.go
```diff
@@ -0,0 +1,146 @@
...
+	var patient fhir.Patient
+	err = s.fhirClient.Read("Patient/"+id, &patient, fhirclient.ResponseHeaders(headers))
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the CarePlan for which the Patient is the subject, get the CareTeams associated with the CarePlan
+	// The search for CarePlan already checks auth for CareTeam, so if we get back a valid CarePlan, we can assume the user has access to the patient
+	// If the user is not part of any of these CareTeams, the search for CarePlan will return an error
+	bundle, err := s.handleSearchCarePlan(ctx, url.Values{"subject-identifier": []string{patientBSNFromIdentifierList(patient.Identifier)}}, headers)
```
reinkrul, line 33: I see creation of the custom search parameter `subject-identifier` often failing, we need to check if this works on Azure FHIR.

orchestrator/careplanservice/handle_getpatient.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func patientBSNFromIdentifierList(identifier []fhir.Identifier) string {
+	for _, id := range identifier {
+		if id.System != nil && *id.System == "http://fhir.nl/fhir/NamingSystem/bsn" {
+			return fmt.Sprintf("%s|%s", *id.System, *id.Value)
+		}
+	}
+	return ""
+}
```
reinkrul, line 146: Do we just support BSNs, or any Patient identifier? Since it's a Dutch number (maybe ask Bram)?

orchestrator/careplanservice/handle_getpatient.go
reinkrul, line 33: shouldn't there be a function similar to `handleTaskBasedResourceAuth`?

orchestrator/careplanservice/handle_getpatient.go
```diff
@@ -0,0 +1,146 @@
...
+
+	patientRefs := make([]string, 0)
+
+	// Iterate through each CareTeam to see if the requester is a participant, if not, remove any patients from the bundle that are part of the CareTeam
+	for _, cp := range carePlans {
+		var ct fhir.CareTeam
+		for _, c := range careTeams {
+			if *cp.CareTeam[0].Reference == fmt.Sprintf("CareTeam/%s", *c.Id) {
+				ct = c
+				break
```
reinkrul, line 113: what if t isn'tfound? Is it possible?

orchestrator/careplanservice/handle_getpatient.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+		participant := coolfhir.FindMatchingParticipantInCareTeam([]fhir.CareTeam{ct}, principal.Organization.Identifier)
```
reinkrul, line 116: I don't think `FindMatchingParticipantInCareTeam` takes the start- and end date into account?

orchestrator/careplanservice/handle_getquestionnaire.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	fhirclient "github.com/SanteonNL/go-fhir-client"
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+// handleGetQuestionnaire fetches the requested Questionnaire and validates if the requester is authenticated
+func (s *Service) handleGetQuestionnaire(ctx context.Context, id string, headers *fhirclient.Headers) (*fhir.Questionnaire, error) {
+	// Verify requester is authenticated
```
reinkrul, line 12: I don't think you need an additional check here, since it's handled by the middleware in `service.go` (`auth.Secured` or something)

orchestrator/careplanservice/handle_getquestionnaireresponse.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil, err
+	}
+
+	var questionnaireResponse fhir.QuestionnaireResponse
+	err = s.fhirClient.Read("QuestionnaireResponse/"+id, &questionnaireResponse, fhirclient.ResponseHeaders(headers))
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.handleTaskBasedResourceAuth(ctx, headers, questionnaireResponse.BasedOn, "QuestionnaireResponse")
```
reinkrul, line 24: can you think of situations where allowing updates to `BasedOn` would be a vulnerability?

orchestrator/careplanservice/handle_getservicerequest.go
```diff
@@ -0,0 +1,40 @@
...
+	// If Task validation passes, the user has access to the ServiceRequest
+	bundle, err := s.handleSearchTask(ctx, map[string][]string{"focus": {"ServiceRequest/" + *serviceRequest.Id}}, headers)
+	if err != nil {
+		return nil, err
+	}
+	// If bundle is empty, the user does not have access to the ServiceRequest
+	if len(bundle.Entry) == 0 {
+		return nil, &coolfhir.ErrorWithCode{
+			Message:    "User does not have access to ServiceRequest",
+			StatusCode: http.StatusUnauthorized,
```
reinkrul, line 35: I'd think Forbidden

orchestrator/careplanservice/handle_getservicerequest.go
```diff
@@ -0,0 +1,40 @@
...
+		return nil, err
+	}
+
+	var serviceRequest fhir.ServiceRequest
+	err = s.fhirClient.Read("ServiceRequest/"+id, &serviceRequest, fhirclient.ResponseHeaders(headers))
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: This query is going to become more expensive as we create more tasks, we should look at setting ServiceRequest.BasedOn or another field to the Task ID
```
reinkrul, line 25: Shouldn't there only be 1 primary Task referring to this ServiceRequest?

reinkrul (changes requested):
I think this is a better approach. Still complicated, we might be able to simplify it further in future. Can you double-check that you have negative test cases for those cases when access should be denied?

I think the end-to-end tests should ultimately do the same authorization checks, since it's all to easy to forget or remove a single very important test case. Better have a similar test in a different test suite in that case (robust systems have redundancy built in to avoid crashing down if 1 component fails).

orchestrator/careplanservice/handle_getcondition.go
dwelman: Yes good point

orchestrator/careplanservice/handle_getcondition.go
dwelman: Ah yes fair point, don't want the requester to know this. I will log and return the same error as earlier

orchestrator/careplanservice/handle_getpatient.go
dwelman: It is possible, if not found it will pass an empty CareTeam to `FindMatchingParticipantInCareTeam`, which will return nil.

I can add a more explicit check if you'd like but it is already covered

orchestrator/careplanservice/handle_getpatient.go
dwelman: It does not. In the past we had some use case required for (I think) task updates which required the participant to be a member of the team but didn't need to be active. 

We do have a method that validates they are active so I will add that and add additional test cases

orchestrator/careplanservice/handle_getquestionnaire.go
dwelman: Ah yes it is. I'll remove the check from the methods that don't explicitly need the principal

orchestrator/careplanservice/handle_getquestionnaireresponse.go
dwelman: I can't think of situations where we would want to allow updates to BasedOn in the first place, shouldn't this be set during creation and left as immutable?

orchestrator/careplanservice/handle_getservicerequest.go
dwelman: Correct. Always get these mixed up

orchestrator/careplanservice/handle_get_test.go
```diff
@@ -0,0 +1,727 @@
...
+			returnedResource: &patient1,
+		},
+	}
+	questionnaireTests := []TestStruct[fhir.Questionnaire]{
+		{
+			ctx:          context.Background(),
+			name:         "No auth",
+			id:           "1",
+			resourceType: "Questionnaire",
+			expectError:  true,
```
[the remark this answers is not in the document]
dwelman: Yes that is a better idea. I copied these over from much older tests and didn't think about changing (tunnel vision), I will assert the actual error we expect to make the tests more robust

orchestrator/careplanservice/handle_getpatient.go
dwelman: Hmm.. That's a good point. All our resources only use BSNs but let me confirm with Bram

orchestrator/careplanservice/handle_getpatient.go
dwelman: Good idea, will implement

dwelman: I can see value in that in the future when we have more examples of resources validated by CarePlan, but at the moment we are only using this for Patient so I don't see much value in it at this time

orchestrator/careplanservice/handle_getpatient.go
bramwesselo: We should support any patient-identifier. Patient-instances of SAZ have lot's of identifiers that they might use for updates/reads (e.g. ziekenhuis-patientnummer, EPIC-FHIR-id, etc.)

dwelman: Noted, thanks. Will make this more flexible

orchestrator/careplanservice/handle_getpatient.go
bramwesselo: I would de-prioritize this one. I don't think we need this custom search-param as we now use 'normal' local references to a patient-instance. The (CarePlan) 'subject' or 'patient' search-param is a standard one for HAPI and Azure FHIR.

orchestrator/careplanservice/handle_get_test.go
dwelman: I have refactored the tests into their own methods and simplified the unit test cases a bit, it should be cleaner and easier to look over now

orchestrator/careplanservice/handle_getpatient.go
dwelman: Ok, noted. I will switch to searching on Subject or Patient (need to investigate which we use)

orchestrator/careplanservice/handle_getpatient.go
dwelman: Discussed with Bram, access requirements for these resources specifically is that any team member with access should be able to view these resources, they do not need to be an active participant

orchestrator/careplanservice/handle_getservicerequest.go
dwelman: Ahh that's a fair point and good to keep in mind. It will be good to cache in the future but yeah not a concern in the meantime

orchestrator/careplanservice/handle_get_test.go
```diff
@@ -0,0 +1,727 @@
...
+	_ = json.Unmarshal(carePlan1Raw, &carePlan1)
+	careTeam2Raw, _ := os.ReadFile("./testdata/careteam-2.json")
+	var careTeam2 fhir.CareTeam
+	_ = json.Unmarshal(careTeam2Raw, &careTeam2)
+
+	patient1Raw, _ := os.ReadFile("./testdata/patient-1.json")
+	var patient1 fhir.Patient
+	_ = json.Unmarshal(patient1Raw, &patient1)
+
+	taskTests := []TestStruct[fhir.Task]{
```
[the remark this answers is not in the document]
dwelman: They have been split

orchestrator/careplanservice/handle_getpatient.go
reinkrul: > I would de-prioritize this one. I don't think we need this custom search-param as we now use 'normal' local references to a patient-instance. The (CarePlan) 'subject' or 'patient' search-param is a standard one for HAPI and Azure FHIR.

Yes, but...: we use a logical identifier in `CarePlan.for` with BSN, so we still need to search for the patient using logical identifier. I can't remember why @JorisHeadease decided to make a custom search parameter.. 🤔

reinkrul (approved):

PR #139, opened 2024-11-15
INT-419: Microsoft Azure FHIR compatiblity

reinkrul:
Note: ongoing tests, more might be added.

We found issues with the processing of `POST Task` transaction result processing on Azure FHIR.

Changes:
- Log transaction request and result bundle
- Introduce ORCA_LOGLEVEL config parameter (default: info)
- Support Azure FHIR's absolute URL (HAPI provides relative URLs) in bundle response `Location` header, which is used to select the right resource to return from a TX bundle

TODO:
- [x] Set tx logging to TRACE
- [x] Check what log level is enabled by default

### DIFF
--- a/orchestrator/careplanservice/handle_createtask.go
+++ b/orchestrator/careplanservice/handle_createtask.go
@@ -190,7 +190,21 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 	return func(txResult *fhir.Bundle) (*fhir.BundleEntry, []any, error) {
 		var createdTask fhir.Task
 		result, err := coolfhir.FetchBundleEntry(s.fhirClient, txResult, func(_ int, entry fhir.BundleEntry) bool {
-			return entry.Response.Location != nil && strings.HasPrefix(*entry.Response.Location, "Task/")
+			if entry.Response.Location == nil {
+				return false
+			}
+			// HAPI uses relative Location URLs, Microsoft Azure FHIR uses absolute URLs.
+			createdResourceLocation := *entry.Response.Location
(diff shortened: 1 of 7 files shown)


orchestrator/careplanservice/handle_createtask.go
```diff
@@ -190,7 +190,21 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
-			return entry.Response.Location != nil && strings.HasPrefix(*entry.Response.Location, "Task/")
+			if entry.Response.Location == nil {
+				return false
+			}
+			// HAPI uses relative Location URLs, Microsoft Azure FHIR uses absolute URLs.
+			createdResourceLocation := *entry.Response.Location
+			createdResourceLocation = strings.TrimPrefix(createdResourceLocation, s.fhirURL.String())
+			// depending on the base URL ending with slash or not, we might end up with a leading slash.
+			// Trim it for deterministic comparison.
+			createdResourceLocation = strings.TrimPrefix(createdResourceLocation, "/")
```
dwelman, line 201: Why not handle this in one trim with `s.fhirURL.String()+"/"` ?

dwelman (approved):
Minor nitpick but otherwise looks good

orchestrator/careplanservice/handle_createtask.go
tader: Say some FHIR service would return a local reference like `/Task/2`, in that case it would not strip the leading `/` by stripping the base url of the service including the `/`

tader (commented):
🚀

orchestrator/careplanservice/handle_createtask.go
dwelman: Ahhhhh yes you're right, fair point

PR #143, opened 2024-11-19
Reference Patient resource in Task

tader:


frontend/lib/fhirUtils.ts
```diff
@@ -93,7 +93,9 @@ export const getTask = (serviceRequest: ServiceRequest, primaryCondition: Condit
 
         //TODO: Not setting this, made by CPS - but should maybe set it for existing
         basedOn: carePlanRef && [carePlanRef],
-        for: serviceRequest.subject,
+        for: {
+            reference: patient.id
+        },
```
JorisHeadease, line 98: The CPS won't understand this relative reference, as it doesn't have the `Patient` resource, and it isn't sent. It might work due to not having referential integrity checks enabled. 

I think this function should be rewritten so that it sends out a `Bundle` to the CPS that is according to the spec: https://santeonnl.github.io/shared-care-planning/Bundle-cps-bundle-01.json

So in this case, the ServiceRequest also doesn't have to be "cleaned up", but all relative reference need to be replaced with bundle references. This way, the CPS can copy the resources, and assign their own `id` values.

JorisHeadease (changes requested):

frontend/lib/fhirUtils.ts
tader: I already had a feeling this would be too easy ;)

frontend/lib/fhirUtils.ts
```diff
@@ -77,8 +77,40 @@ export const getCarePlan = (patient: Patient, conditions: Condition[], carePlanN
...
+        resourceType: "Bundle",
+        type: "transaction",
+        entry: [
+            {
+                fullUrl: "urn:uuid:patient",
+                resource: {...patient, id: undefined},
+                request: {
+                    method: "PUT",
+                    url: `Patient?identfier=http://fhir.nl/fhir/NamingSystem/bsn|${getBsn(patient)}`
+                }
```
jscharp, line 91: I think it's probably safer to only create the `Patient` resource if it's new to the CPS. If it already knows the `Patient`, do not update it. Can also solve potential ETag/422 errors in certain FHIR servers. Something like this:

```suggestion
                request: {
                    method: "POST",
                    url: "Patient",
                    ifNoneExist: `identifier=http://fhir.nl/fhir/NamingSystem/bsn|${getBsn(patient)}`,
                }
```

jscharp (commented):
Looks good @tader! Added 1 suggestion that might prevent future issues.

JorisHeadease (changes requested):

JorisHeadease (approved):
LGTM @tader! There is still @reinkrul's comment, but that could also be resolved in a separate PR.

PR #147, opened 2024-11-19
[FIX] Implement task-output search param for QuestionnaireResponse auth

dwelman:
A user that has access to a Task also has access to any QuestionnaireResponse in its Output


reinkrul:
I hope the resources get re-indexed by Azure FHIR when a new search parameter is created...

bramwesselo:
We must add a step to reindex our custom searchparameter:
(source: https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/how-to-run-a-reindex )

`POST {{FHIR_URL}}/$reindex 
content-type: application/fhir+json
{ 
"resourceType": "Parameters",  
"parameter": [
    {
      "name": "targetSearchParameterTypes",
      "valueString": "{url of custom search parameter. In case of multiple custom search parameters, url list can be comma separated.}"
    }
] 
}`

PR #148, opened 2024-11-20
INT-438: Sanitize logged URLs

reinkrul:


reinkrul:
Probably, the better approach is to reconfigure it on the logger instead. But, zerolog doesn't support adjusting the formatted message. Made an issue for it, to check whether it's a feasible feature: https://github.com/rs/zerolog/issues/697

PR #154, opened 2024-11-20
Persist Zorgplatform workflow-id

JorisHeadease:
Mapping the `ServiceRequest` to the Zorgplatform `workflow-id`. Zorgplatform does not work with a `ServiceRequest`, but a `Task` for orders. They serve the same purpose, so they are mapped. This allows for querying the Zorgplatform data later on.

### DIFF
--- a/orchestrator/careplancontributor/applaunch/zorgplatform/service.go
+++ b/orchestrator/careplancontributor/applaunch/zorgplatform/service.go
@@ -377,6 +377,12 @@ func (s *Service) getSessionData(ctx context.Context, accessToken string, launch
 
 	serviceRequest := &fhir.ServiceRequest{
 		Status: fhir.RequestStatusActive,
+		Identifier: []fhir.Identifier{
+			{
+				System: to.Ptr("https://api.zorgplatform.online/fhir/v1"),
+				Value:  to.Ptr(launchContext.WorkflowId),
+			},
+		},
 		Code: &fhir.CodeableConcept{
 			Coding: []fhir.Coding{
 				// Hardcoded, we only do Telemonitoring for now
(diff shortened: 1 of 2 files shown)


orchestrator/careplancontributor/applaunch/zorgplatform/service_test.go
```diff
@@ -227,6 +228,18 @@ func TestService(t *testing.T) {
 			serviceRequestRef := sessionData.StringValues["serviceRequest"]
 			require.NotEmpty(t, serviceRequestRef)
 			require.IsType(t, fhir.ServiceRequest{}, sessionData.OtherValues[serviceRequestRef])
+			t.Run("check Workflow-ID identifier is properly set on the ServiceRequest", func(t *testing.T) {
```
reinkrul, line 232: I simplified the test assertion

orchestrator/careplancontributor/applaunch/zorgplatform/service.go
```diff
@@ -377,6 +377,12 @@ func (s *Service) getSessionData(ctx context.Context, accessToken string, launch
 
 	serviceRequest := &fhir.ServiceRequest{
 		Status: fhir.RequestStatusActive,
+		Identifier: []fhir.Identifier{
+			{
+				System: to.Ptr("https://api.zorgplatform.online/fhir/v1/Task"),
```
reinkrul, line 382: This system looks a lot like an actual URL of the Zorgplatform workflow API. Is that practical? Might cause confusion?

reinkrul (approved):
Comment about the namespace

orchestrator/careplancontributor/applaunch/zorgplatform/service.go
JorisHeadease: It could, perhaps when using agains ZPF ACC. Used the value that was designed in the ticket, I think it should be okay as the system is "only" a namespace. It does clearly map to a workflow-id Task, a more generic value could be something like `https://zorgplatform.online/workflows`

PR #157, opened 2024-11-21
CPS: return original bundle stripped from unauthorized resources

reinkrul:
To prevent accidentally omitting fields.

The problem was we created a new Bundle, now we simply take a copy instead.

### DIFF
--- a/orchestrator/careplanservice/util.go
+++ b/orchestrator/careplanservice/util.go
@@ -148,24 +148,19 @@ func matchResourceIDs(request *FHIRHandlerRequest, idFromResource *string) error
 }
 
 // filterMatchingResourcesInBundle will find all resources in the bundle of the given type with a matching ID and return a new bundle with only those resources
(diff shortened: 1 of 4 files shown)


orchestrator/careplanservice/util.go
```diff
@@ -148,48 +148,40 @@ func matchResourceIDs(request *FHIRHandlerRequest, idFromResource *string) error
...
+					newBundle.Entry[j] = entry
+					j++
 					break
 				}
 			}
 		}
 	}
-	newBundle.Entry = append(newBundle.Entry, operationOutcomeErrors...)
-
+	newBundle.Entry = newBundle.Entry[:j]
```
dwelman, line 185: Ah `[:j]`, we meet again :P

reinkrul: I wrote this, specially, for you

PR #160, opened 2024-11-25
Zorgplatform: add search path for existing Task to show if existing Tasks

reinkrul:
Also update the Zorgplatform logical identifier for Task to Zorgplatform's Workflow ID XML namespace.


JorisHeadease:
So this is where the system change was at @reinkrul! I've merged main into the branch an focused on the system. The PR was laying around a bit due to changes in approach (workflow-id on the ServiceRequest). So it mainly just changes the ZPF system now..

JorisHeadease (approved):

PR #161, opened 2024-11-25
Troubleshoot Bad Request response without response body

reinkrul:
On Azure FHIR, the following query returns HTTP 400 Bad Request, without response body:

```
/orca/cpc/cps/fhir/ServiceRequest?identifier=2.16.528.1.1007.3.3.21514.ehr.orders%7C99534756439a
```

The same goes for `ServiceRequest?_id=2`

Fix:

I couldn't find out why it failed, but it was something in the usage of `httputil.ReverseProxy`. I rewrote our FHIR proxy to use `fhirclient.Client`, which worked. This should also improve the consistency of HTTP responses (headers, marshalling format). This thus applies to the actual issue (CPS proxying to Azure FHIR), but is now also used by the CPC->CPS proxy (which uses the same). But tests still work, so should be OK.

Also improved proxy logging (it now logs a proxy name, since some requests are routed through 2 proxies, making logging very hard to understand) and fixed a bug in the FHIR error response marshaller (sometimes an OperationOutcome error wasn't seen as such, depending on whether a pointer to it was passed).

Note that, as of consequence, we don't support anything other than POST, GET and PUT, but I don't think we need to, for now (since we don't support it anyways in ORCA).

### DIFF
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -353,8 +353,6 @@ func (s *Service) handleSearch(httpRequest *http.Request, httpResponse http.Resp
 		bundle, err = s.handleSearchTask(httpRequest.Context(), httpRequest.URL.Query(), headers)
 	case "Patient":
 		bundle, err = s.handleSearchPatient(httpRequest.Context(), httpRequest.URL.Query(), headers)
-	//case "ServiceRequest":
-	//	_, bundle, err = handleSearchResource[fhir.Patient](s, "ServiceRequest", httpRequest.URL.Query(), headers)
 	default:
 		log.Warn().Ctx(httpRequest.Context()).
 			Msgf("Unmanaged FHIR operation at CarePlanService: %s %s", httpRequest.Method, httpRequest.URL.String())
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -136,7 +136,7 @@ func (r FHIRHandlerRequest) bundleEntry() fhir.BundleEntry {
 type FHIRHandlerResult func(txResult *fhir.Bundle) (*fhir.BundleEntry, []any, error)
 
 func (s *Service) RegisterHandlers(mux *http.ServeMux) {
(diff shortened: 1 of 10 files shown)


reinkrul:
I couldn't find out why it failed, but it was something in the usage of `httputil.ReverseProxy`. I rewrote our FHIR proxy to use `fhirclient.Client`, which worked. This should also improve the consistency of HTTP responses (headers, marshalling format).

orchestrator/careplanservice/service.go
```diff
@@ -354,6 +353,8 @@ func (s *Service) handleSearch(httpRequest *http.Request, httpResponse http.Resp
 		bundle, err = s.handleSearchTask(httpRequest.Context(), httpRequest.URL.Query(), headers)
 	case "Patient":
 		bundle, err = s.handleSearchPatient(httpRequest.Context(), httpRequest.URL.Query(), headers)
+	//case "ServiceRequest":
```
dwelman, line 356: Can this be removed?

orchestrator/lib/coolfhir/error.go
```diff
@@ -97,15 +97,15 @@ func WriteOperationOutcomeFromError(err error, desc string, httpResponse http.Re
...
+	if errors.As(err, operationOutcomeErr) || errors.As(err, &operationOutcomeErr) {
 		if operationOutcomeErr.HttpStatusCode > 0 {
 			statusCode = operationOutcomeErr.HttpStatusCode
 		}
 		operationOutcome = operationOutcomeErr.OperationOutcome
 	} else {
 		// Error type: ErrorWithCode
-		var errorWithCode *ErrorWithCode
+		var errorWithCode = new(ErrorWithCode)
 		if errors.As(err, &errorWithCode) {
```
dwelman, line 109:
```suggestion
		if errors.As(err, errorWithCode) || errors.As(err, &errorWithCode) {
```
For parity with the above check

dwelman (approved):
Looks good, minor comments

PR #162, opened 2024-11-26
[INT-460] Only create SearchParameter if not exists

dwelman:


orchestrator/careplanservice/service.go
```diff
@@ -571,12 +571,23 @@ func (s *Service) ensureSearchParameterExists(ctx context.Context) {
 			req.Header.Set("If-None-Exist", fmt.Sprintf("url=%s", param.SearchParam.Url))
```
reinkrul, line 571: do we still need this, then?

dwelman: Probably not, will confirm

reinkrul (approved):

PR #164, opened 2024-11-26
INT-399: Reject Task if Task Filler engine can't execute it

reinkrul:
Needs to implement the following state diagram: https://github.com/SanteonNL/shared-care-planning/blob/prestation-hl7-wgm/input/images-source/notification-task-flow.svg


dwelman (approved):
Looks good, unit test needs to be fixed before merge

dwelman (approved):

PR #167, opened 2024-11-28
INT-464: Prepare fetching Questionnaires from FHIR API

reinkrul:
Introduces an interface for `WorkflowProvider`, which we can implement with a store fetching Questionnaires from a FHIR API.

### DIFF
--- a/orchestrator/careplancontributor/handle_taskfiller.go
+++ b/orchestrator/careplancontributor/handle_taskfiller.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"slices"
+	"github.com/SanteonNL/orca/orchestrator/lib/slices"
 	"strings"
 
 	fhirclient "github.com/SanteonNL/go-fhir-client"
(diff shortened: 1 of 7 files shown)


orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -320,30 +299,46 @@ func (s *Service) selectWorkflow(cpsClient fhirclient.Client, task *fhir.Task) (
...
 		}
 	}
-	// find matching workflow
-	var matchedCodes []string
-	var matchedWorkflow *taskengine.Workflow
-	for _, coding := range candidateCodes {
-		key := fmt.Sprintf("%s|%s", *coding.System, *coding.Code)
-		if slices.Contains(matchedCodes, key) {
-			// duplicate code, doesn't matter
+	taskReasonCodes = slices.Deduplicate(taskReasonCodes, func(a, b fhir.Coding) bool {
```
dwelman, line 314: It seems like the check above should ensure duplicates are not added to the list, if we are going to run the dedupe anyway why not add all the matching codes and skip the comparison?

reinkrul: Then I'd have to write it twice (for task.reasoncode and task.reasonReference.code), and it's a bit of an elaborate Task (checking system and code aren't nil, then comparing both). So that's why it's cognitivially easier this way.

dwelman: Fine with me

dwelman (approved):

PR #168, opened 2024-11-28
INT-464: Determine Task Engine workflow from FHIR HealthcareService and Questionnaires

reinkrul:
This PR switches the Task Engine's embedded Questionnaire lookup to the configured FHIR API:
- For validating the `ServiceRequest.code, Condition.code` combination, it checks if there's an associated `HealthcareService`.
- The questionnaire is then loaded by searching for the coding
The embedded questionnaires are now for testing purposes only.

Since this creates a new issue for managing these resources, ORCA now reads FHIR Bundles from configured URLs and loads them into the FHIR API. For security reasons, it refuses bundles that contain resources other than Questionnaire and HealthcareService.

Random improvements:
- e2e test had did:nuts enabled, changed this to did:web only
- e2e test mounted Nuts config files into orchestrator, removed this

TODO:
- [x] Configure local dev setup
- [x] Configure test/dev deployments
- [x] Fix e2e tests


orchestrator/careplancontributor/service.go
```diff
@@ -40,11 +40,37 @@ func New(
...
+			return nil, err
+		}
+	}
+	// Load Questionnaire-related resources for the Task Filler Engine from the configured URLs into the Questionnaire FHIR API
+	go func(ctx context.Context) {
+		if len(config.TaskFiller.QuestionnaireSyncURLs) > 0 {
+			log.Info().Ctx(ctx).Msgf("Synchronizing Task Filler Questionnaires resources from %d URLs", len(config.TaskFiller.QuestionnaireSyncURLs))
+			for _, u := range config.TaskFiller.QuestionnaireSyncURLs {
+				if err := coolfhir.ImportResources(ctx, questionnaireFhirClient, []string{"Questionnaire", "HealthcareService"}, u); err != nil {
+					log.Error().Ctx(ctx).Err(err).Msgf("Failed to synchronize Task Filler Questionnaire resources (url=%s)", u)
```
dwelman, line 66: I'm wondering, will this cause a deeper problem later if a flow tries to use a Questionnaire that could not be synced? 

Not to say we should crash the entire service if one fails to load, just thinking of edge cases.

I know it will fail out if the request comes in with an unsupported questionnaire, but it feels like a separate issue if a request fails for a questionnaire that should be supported but couldn't be loaded

orchestrator/lib/coolfhir/import.go
```diff
@@ -0,0 +1,89 @@
...
+		if httpResponse.StatusCode <= 199 || httpResponse.StatusCode >= 300 {
+			return fmt.Errorf("unexpected status code: %d", httpResponse.StatusCode)
+		}
+		bundleData, err = io.ReadAll(io.LimitReader(httpResponse.Body, 1024*1024*10+1)) // Limit to 10mb
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO: Might want to make a "secure" http.Client that always does this
```
dwelman, line 50: Further in the vein of security, do we maybe want to provide some kind of whitelist for URLs we can download resources from?

dwelman (approved):
Approved with non-blocking comments

orchestrator/lib/coolfhir/import.go
reinkrul: I considered this;
- The configured URLs are part of deployment configuration so at least 2 people should be involved
- You should use HTTPS URLs to avoid a MITM changing stuff
- The risk is probably very small, only frustrating the process.

I limit the resources being loaded into the FHIR server to Questionnaire and HealthcareService resources (and at 10mb), so I think the risk is quite low.

dwelman: Great, happy with that. Nice consideration to make

orchestrator/careplancontributor/service.go
reinkrul: Well, it only needs to succeed once (until you update it of course). 

It could be a fact on the service health page? (which we don't have...). I added this as a thing we could implement as part of our monitoring efforts.

PR #169, opened 2024-11-29
Enrollment UI rework

JorisHeadease:
Initial version of the UI rework. There will be some more changes required in the future, but this PR will offer the foundation.

Also contains a major upgrade from nextjs 14 to 15. And also note that a lot of the styling changes are embedded in the questionnaire itself.


frontend/lib/store/enrollment-store.ts
```diff
@@ -76,9 +76,9 @@ const fetchLaunchContext = async (set: (partial: StoreState | Partial<StoreState
     } else {
         //TODO: We can remove this when going live, this is useful during development
         launchContext = {
-            "patient": "Patient/2",
-            "serviceRequest": "ServiceRequest/4",
-            "practitioner": "Practitioner/7"
+            "patient": "Patient/4",
+            "serviceRequest": "ServiceRequest/6",
+            "practitioner": "Practitioner/8"
```
reinkrul, line 81: I'd like to remove this sooner, rather than later. Why do we need this?

reinkrul (changes requested):

dwelman (approved):

frontend/lib/store/enrollment-store.ts
JorisHeadease: It's mainly so I can test my changes in a few seconds ([Fast Refresh](https://nextjs.org/docs/architecture/fast-refresh)) on the dev server. These values map onto the first created records when using the dev server. The front-end can be started next to it, in a dev server, and just use that data. Otherwise you need to keep building docker images and test it via the dev deployment.

reinkrul: I'll approve if you make an issue for it on the backlog

reinkrul (approved):

frontend/lib/store/enrollment-store.ts
JorisHeadease: https://linear.app/zorg-bij-jou/issue/INT-471/orca-or-remove-the-dev-server-launch-context

PR #176, opened 2024-12-09
INT-478: Update outdated dependencies

reinkrul:
Also remove smartonfhir backend adapter build; we don't use it, so only slows down the build and generates artifacts no-one uses (so wasted CPU cycles and bytes).


reinkrul:
@JorisHeadease can you update the `micromatch` package in `viewer_simulator`? My knowledge of npm is lacking here... (`npm audit` didn't find the CVC).

PR #217, opened 2024-12-11
INT-496: ORCA CPC ZbJ | BUG: The questionnaire is created multiple times in the FHIR store, causing the task submission to fail.

rolandgroen:
Use "Update" for Questionnaire creation in transaction.

Replaced `Create` with `Update` for the Questionnaire in the FHIR transaction, aligning with the correct resource reference format. Adjusted the `questionnaireRef` to use "Questionnaire/" prefix for consistency with FHIR standards. Removed obsolete comment and unnecessary entry modifications.

### DIFF
--- a/orchestrator/careplancontributor/handle_taskfiller.go
+++ b/orchestrator/careplancontributor/handle_taskfiller.go
@@ -268,14 +268,12 @@ func (s *Service) createSubTaskOrAcceptPrimaryTask(ctx context.Context, cpsClien
 	}
 
 	// Create a new SubTask based on the Questionnaire reference
-	questionnaireRef := "urn:uuid:" + *questionnaire.Id
+	questionnaireRef := "Questionnaire/" + *questionnaire.Id
 	subtask := s.getSubTask(primaryTask, questionnaireRef)
 	subtaskRef := "urn:uuid:" + *subtask.Id
 
 	tx := coolfhir.Transaction().
-		Create(questionnaire, coolfhir.WithFullUrl(questionnaireRef), func(entry *fhir.BundleEntry) {
-			entry.Request.Url = "Questionnaire" // TODO: remove this after changed to fhir.Questionnaire
-		}).
+		Update(questionnaire, questionnaireRef).
 		Create(subtask, coolfhir.WithFullUrl(subtaskRef))
 
 	bundle := tx.Bundle()


orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -268,14 +268,12 @@ func (s *Service) createSubTaskOrAcceptPrimaryTask(ctx context.Context, cpsClien
...
-	questionnaireRef := "urn:uuid:" + *questionnaire.Id
+	questionnaireRef := "Questionnaire/" + *questionnaire.Id
 	subtask := s.getSubTask(primaryTask, questionnaireRef)
 	subtaskRef := "urn:uuid:" + *subtask.Id
 
 	tx := coolfhir.Transaction().
-		Create(questionnaire, coolfhir.WithFullUrl(questionnaireRef), func(entry *fhir.BundleEntry) {
-			entry.Request.Url = "Questionnaire" // TODO: remove this after changed to fhir.Questionnaire
-		}).
+		Update(questionnaire, questionnaireRef).
```
reinkrul, line 276: I very much wonder whether this fixes the issue, since CPS doesn't allow setting the ID of the created FHIR resource (if it does, it's a bug in CPS). So the ID of the created FHIR resources won't be equal, so the update will always be an upsert which creates the resource.

Better to use upsert with a logical identifier here?

reinkrul (changes requested):

orchestrator/careplancontributor/handle_taskfiller.go
rolandgroen: It fixes the issue given the second Questionnaire is not created. 

I don't know how to do an upsert with coolfhir.

reinkrul (approved):

orchestrator/careplancontributor/handle_taskfiller.go
rolandgroen: On how it works, I think that is because it has become an update. 

I'm not sure if it's worth the time figuring out the conditional create works vs. the fact that the submission flow is broken, and the basic UC cannot be performed. 

I propose sticking to this solution, as you said: this code is being poked around soon anyway.

reinkrul:
Does this bug only happen on HAPI? Since after deploying to test and doing enrolment multiple times, this bug doesn't show up there.

PR #219, opened 2024-12-12
INT-499: Refactor token type structure to use a function for Subject

JorisHeadease:
Added tests to validate differences in token types, and made sure the integration test also tests both token types

### DIFF
--- a/orchestrator/careplancontributor/applaunch/zorgplatform/handle_create_access_token.go
+++ b/orchestrator/careplancontributor/applaunch/zorgplatform/handle_create_access_token.go
@@ -25,16 +25,15 @@ type SecureTokenService interface {
 }
 
 type TokenType struct {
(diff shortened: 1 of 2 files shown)


dwelman (approved):

orchestrator/careplancontributor/applaunch/zorgplatform/handle_create_access_token.go
```diff
@@ -25,16 +25,15 @@ type SecureTokenService interface {
 }
 
 type TokenType struct {
-	Type         string
+	Subject      func(element *etree.Element, launchContext *LaunchContext, applicationIssuer string)
```
reinkrul, line 28: it's not nice, but I can't think of a nicer solution

reinkrul (approved):

PR #228, opened 2024-12-13
INT-498 | Fix CPC unit tests

dwelman:
Implement mocked reverse proxy to re-enable unit and integration tests in CPC


orchestrator/mocks/reverse_proxy_mock.go
```diff
@@ -0,0 +1,32 @@
...
+	"net/http"
+	"net/url"
+)
+
+// MockReverseProxy is a mock implementation of a ReverseProxy for unit tests.
+type MockReverseProxy struct {
+	Proxy coolfhir.HttpProxy
+}
+
+func NewMockReverseProxy(target *url.URL, orcaPublicUrl *url.URL, transport http.RoundTripper, allowCaching bool) (*MockReverseProxy, error) {
```
reinkrul, line 15: `coolfhir.HttpProxy` implements `ReverseProxy`, so this type only exists for the defaults below?

reinkrul (commented):
see comments

orchestrator/mocks/reverse_proxy_mock.go
dwelman: Yeah I tried to get this to work across the unit and integ test before realising I shouldn't use it for the integ test, and I think based on how I set it up for the integ test I might be able to do similar for the unit test. Will check

dwelman: Yeah this was leftovers from when it was genuinely a mock client and not just a wrapper, I have removed it

PR #229, opened 2024-12-13
INT-464: Implement in-memory workflow provider 

reinkrul:
As alternative to Questionnaire/HealthcareService FHIR resources from FHIR API. Useful if a vendor doesn't want to load questionnaires/healthcareservices in their FHIR instance.

It filters the HealthcareService and Questionnaire v.s. the ServiceRequest code and Condition code in-memory, instead of searching them in the FHIR API.

The downside is, that on boot the (external) source of the FHIR bundles MUST be up and running, otherwise ORCA will fail to start.

### DIFF
--- a/orchestrator/careplancontributor/taskengine/workflow.go
+++ b/orchestrator/careplancontributor/taskengine/workflow.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	fhirclient "github.com/SanteonNL/go-fhir-client"
+	"github.com/SanteonNL/orca/orchestrator/lib/coolfhir"
 	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+	"net/http"
+	"net/url"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


orchestrator/careplancontributor/taskengine/workflow.go
```diff
@@ -82,6 +85,93 @@ func (f FhirApiWorkflowProvider) QuestionnaireLoader() QuestionnaireLoader {
...
+
+	var healthcareServices []fhir.HealthcareService
+	if err := coolfhir.ResourcesInBundle(&bundle, coolfhir.EntryIsOfType("HealthcareService"), &healthcareServices); err != nil {
+		return fmt.Errorf("could not extract healthcare services from bundle: %w", err)
+	}
+	e.healthcareServices = append(e.healthcareServices, healthcareServices...)
+	return nil
+}
+
+func (e *MemoryWorkflowProvider) Provide(ctx context.Context, serviceCode fhir.Coding, conditionCode fhir.Coding) (*Workflow, error) {
```
dwelman, line 123: With `LoadBundle`, `Provide`, and `Load`, it's not the easiest to determine what does what. Maybe the names could be a bit more descriptive?

dwelman (approved):
Looks good, just a comment on naming

orchestrator/careplancontributor/taskengine/workflow.go
reinkrul: Provide and Load are from the interfaces, LoadBundle for provisioning the resources required by this type. I'll some godoc.

PR #238, opened 2024-12-16
INT-502: enrollment UI improvements

JorisHeadease:
- Contains feedback fixes from Juliet
- Fix weird reloading bug of the enrollment confirm screen


dwelman (approved):
I have nothing meaningful to add, I trust that this is a UI improvement :P

PR #253, opened 2024-12-18
INT-503: Cache X-SCP-Context mapping to Zorgplatform Workflow ID

JorisHeadease:
Made a start on caching X-Scp-Content to the ZPF workflow-id and the workflow-id to the SAML token. The latter uses an in-memory TTL cache, so the record is removed when the token expires. Could be changed into a distributed cache when orca needs to scale. TTL record still has to be added, and the unit tests have to be fixed for the map, and implemented for the TTL cache.


reinkrul:
@JorisHeadease can you review this?

PR #255, opened 2024-12-18
INT-511: FHIR proxy ignores upstream errors on POST-search

reinkrul:
A new `err` var was used that shadowed the shared `err` var, which caused the proxy to return status code `0` (which causes issues downstream).


dwelman (approved):

dwelman:
Nice catch!

PR #258, opened 2024-12-19
INT-517: Smaller image for Viewer Demo app

reinkrul:


jdksloan (approved):

JorisHeadease (approved):
Great, this is working now! Before, the standalone had troubles serving the application on the basePath. This caused it not to work with our reverse proxy. This seems to be resolved now, perhaps with the nextjs major upgrade to 15.

PR #280, opened 2025-01-09
INT-527: only map test BSN in strict mode, add graceful shutdown

reinkrul:
To test the `globals` strict mode default setting on server start, I had be able stop the started server. Which got a bit out of hand, since I wanted to support graceful shutdown as well.


orchestrator/careplancontributor/applaunch/zorgplatform/service.go
```diff
@@ -301,8 +301,11 @@ func (s *stsAccessTokenRoundTripper) RoundTrip(httpRequest *http.Request) (*http
 	bsn := *bsnIdentifier.Value
 
 	//TODO: Below is to solve a bug in zorgplatform. The SAML attribute contains bsn "999911120", but the actual patient has bsn "999999151" in the resource/workflow context
-	if bsn == "999911120" {
-		bsn = "999999151"
+	if !globals.StrictMode {
```
jdksloan, line 304: Just for context for me can you explain why we need to this, instead of changing the data in zorgplatform

reinkrul: ChipSoft provides 2 systems we use here:
- HiX, the EHR doctors/nurses at the hospitals for their patient file. This is (as far as I know) an on-premise solution running inside the hospitals
- Zorgplatform, an integration platform for accessing data inside the HiX EHR using FHIR. It allows you to read data (e.g. patient files) from HiX in JSON format and write back to HiX (very limited, actually, I think FHIR flags only).

If you want to create an "app" for HiX (like ORCA allowing hospitals to send orders to other care orgs, like Zorg bij jou), you need to use Zorgplatform, because:
- doesn't run on-premise like HiX
- is usable by multiple hospitals using HiX
- uses FHIR to consume hospital data

Now for this bug: the test environment we use has 2 sets of test data (HiX and Zorgplatform). When a "doctor" launches the ORCA application from HiX, the launch context contains a patient with a BSN (999911120), but this patient has a different BSN in Zorgplatform (BSN 999999151). This a bug/misalignment in the testdata, which prevents testing the ORCA application flow with HiX/Zorgplatform. We didn't have success in getting Chipsoft to fix the testdata.

jdksloan: Got it! Thanks for the great explanation Rein! :)

jdksloan (approved):

PR #297, opened 2025-01-16
INT-473: Added an identical mechanism to view/recover an existing enrolment for the demo launch

JorisHeadease:


orchestrator/go.mod
```diff
@@ -111,6 +111,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
 	github.com/russellhaering/goxmldsig v1.4.0
+	github.com/samply/golang-fhir-models/fhir-models v0.3.2
```
reinkrul, line 114: we already have `github.com/zorgbijjou/golang-fhir-models/fhir-models` which we use everywhere, why need samply here?

orchestrator/lib/coolfhir/util.go
```diff
@@ -399,3 +401,29 @@ func ContainsCoding(code fhir.Coding, concepts ...fhir.CodeableConcept) bool {
 	}
 	return false
 }
+
+func GetTaskByIdentifier(ctx context.Context, fhirClient fhirclient.Client, identifier fhir.Identifier) (*fhir.Task, error) {
```
reinkrul, line 405: would be good to have a unit test. You can use `StubFHIRClient` for easy testing.

reinkrul (changes requested):

dwelman:
No comments from my side, will leave this between you and Rein

orchestrator/lib/coolfhir/util.go
JorisHeadease: Yeah I agree, tried it before, but ran into a circular dependency. It is currently indirectly tested twice in the demo and zpf service. I'll try and fix it later (after I solved the Practitioner logic from ZPF - higher prio issue).

reinkrul (approved):

PR #301, opened 2025-01-16
INT-529: Fix ifNoneExist headers

reinkrul:
TODO:
- [ ] Test if it actually works.


reinkrul:
Not working, we're not passing on the HTTP request headers. Reworking this.

dwelman (approved):

PR #310, opened 2025-01-20
Fixes duplicate URL value and task enrollment failure

dwelman:
Ran into two issues that @rolandgroen also brought up so decided to fix them as it is simple.

One is that enrollment went to a 404 due to the URL having "/new/new" at the end, the other is some leftover change from the ifNotExists revert that was not removed, this has been commented as it was causing enrollment to fail. This issue was picked up on the main branch


reinkrul:
FrontendLandingURL must be configured to `<some base URL>/enrollment`, so this essentially tries to fix broken config. Which doesn't work, since the app launch might need to redirect to `/enrolment`.

reinkrul:
I fixed the config issue (didn't test it, though), and improved the redirect URL building a bit. On test, I saw some double slashes which should now be gone by using `url.JoinPath()` instead of manual concatenation.

reinkrul (approved):

dwelman:
> I fixed the config issue (didn't test it, though), and improved the redirect URL building a bit. On test, I saw some double slashes which should now be gone by using `url.JoinPath()` instead of manual concatenation.

Nice! I must have missed the /new in the config which is why I changed it in code. Thanks for the fix. I'll test and merge

PR #315, opened 2025-01-21
INT-547: Document how ORCA entities map to Zorgplatform and HiX fields

reinkrul:


reinkrul:
Please check whether it's as intended.

PR #316, opened 2025-01-21
INT-522 | Add tests to verify referential integrity when posting bundles

dwelman:
Tests confirm that when a patient and task are created in a bundle with local references, FHIR correctly replaces the references with the new IDs

I could not recreate the issue that was logged in the ticket, however, having these tests ensures functionality stays as expected


e2e-tests/main_test.go
```diff
@@ -350,9 +351,109 @@ func Test_Main(t *testing.T) {
 
 		t.Run("assert Task is rejected", func(t *testing.T) {
 			var rejectedTask fhir.Task
-			err = hospitalOrcaFHIRClient.Read("Task/"+*unsupportedTask.ID, &rejectedTask)
+			err = hospitalOrcaFHIRClient.Read("Task/"+*unsupportedTask.Id, &rejectedTask)
 			require.NoError(t, err)
 			require.Equal(t, fhir.TaskStatusRejected, rejectedTask.Status)
 		})
 	})
+	t.Run("Creating Task and Patient via bundle, supplied fields are persisted, and local references are replaced", func(t *testing.T) {
```
reinkrul, line 359: wouldn't the integration test suffice? This adds quite some burden to the e2e-test, with the orchestrator dependency as well?

dwelman: I wanted to be explicit with what was logged in the ticket but let me see what can be removed, I'll get back to you shortly

dwelman: Oh oops I misread this. The test has been removed

reinkrul (approved):

PR #325, opened 2025-01-24
INT-539: Added the `PractitionerRole` to the Zorgplatform launch. 

JorisHeadease:
This PR contains the following steps:

1. During the Zorgplatform launch, extract the "local userid" and org oid from HiX from the SAML SSO token, and set these as identifiers of a newly created `PractitionerRole`
2. Add this `PractitionerRole` to the session data & the `/context` request in the CPC
3. Modify the frontend, to include the `PractitionerRole` as a `contained` resource on the `Task`. 
4. Modify the viewer to show the PractitionerRole identifiers (if set).

**Note**: This PR required external changes to make `contained` work. See https://github.com/Zorgbijjou/golang-fhir-models/pull/1 for more information.


orchestrator/careplancontributor/applaunch/zorgplatform/handle_validation.go
```diff
@@ -19,13 +19,17 @@ import (
...
+	Bsn              string
+	SubjectNameId    string
+	Practitioner     fhir.Practitioner
+	PractitionerRole fhir.PractitionerRole
+	ServiceRequest   fhir.ServiceRequest
+	WorkflowId       string
 }
 
+const HIX_LOCALUSER_SYSTEM = "https://www.cwz.nl/hix-user"
+const HIX_ORG_OID_SYSTEM = "https://www.cwz.nl/hix-org-oid"
```
reinkrul, line 31: Did we align on this? This adapter is not tenant-specific, but the domain is pretty specific.

reinkrul (changes requested):

orchestrator/careplancontributor/applaunch/zorgplatform/handle_validation.go
reinkrul: (also @bramwesselo)
What I meant, is that CWZ an actual hospital, which can't be hardcoded.

The example we're provided with only has 1 identifier, that of the user. But we're also putting the organization in it. As PractitionerRole, as we have here, is a combination of the 2, shouldn't we have 1 identifier that identifies both?

We could use the domain from the original identifier from ChipSoft (`USER1@2.16.840.1.113883.2.4.3.124.8.50.8`).

Meaning, we could use `2.16.840.1.113883.2.4.3.124.8.50.8` as system, or assigner.

bramwesselo: In this case, the OID is a child under Chipsoft B.V. See printscreen; the '8.50'-suffix is probably for (CS-client-)organizations and the last '.8' probably corresponds to CWZ....
image

I would say we have 2 identifiers here; one to the user (in some namespace) and one to the organization (being an OID, it is already in namespace 2.16.840.1.113883.2.4.3.124)

reinkrul (approved):

PR #326, opened 2025-01-24
INT-559 | Prevent duplicate Task enrollment

dwelman:
Re-enable IfNoneExists headers to ensure enrollment is persisted for the same task per ServiceRequest, prevents duplicate ServiceRequests and Tasks from being created

### DIFF
--- a/frontend/lib/fhirUtils.ts
+++ b/frontend/lib/fhirUtils.ts
@@ -119,7 +119,7 @@ const cleanPatient = (patient: Patient) => {
     return cleanedPatient;
 }
 
-const cleanServiceRequest = (serviceRequest: ServiceRequest, patient: Patient, patientReference: string) => {
+const cleanServiceRequest = (serviceRequest: ServiceRequest, patient: Patient, patientReference: string, taskIdentifier?: string) => {
     // Clean up the ServiceRequest by removing relative references - the CPS won't understand them
     const cleanedServiceRequest = {...serviceRequest, id: undefined};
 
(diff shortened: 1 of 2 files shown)


frontend/lib/fhirUtils.ts
```diff
@@ -207,7 +206,7 @@ export const constructTaskBundle = (serviceRequest: ServiceRequest, primaryCondi
         request: {
             method: "POST",
             url: "ServiceRequest",
-            ifNoneExist: "",
+            ifNoneExist: `ServiceRequest?_id=${serviceRequest.id}`,
```
reinkrul, line 209: The ServiceRequest came from the EHR, and normally the FHIR server assigns the technical ID. Meaning, when the ServiceRequest is created at the CPS, it probably won't have the same ID (although it might work locally on HAPI). We need to use a logical `identifier` here instead.

reinkrul: Also, the resource type (`ServiceRequest`) shouldn't be there, just the search parameters. See https://build.fhir.org/http.html#ccreate

reinkrul (changes requested):

frontend/lib/fhirUtils.ts
dwelman: I have removed the resource types from the queries, coordinating with @JorisHeadease on the correct identifier to use for the ServiceRequest

JorisHeadease: It's still using the `_id` property though. You probably want to conditionally create on the `identifier`. Easiest approach is probably to share the taskIdentifier

```
identifier=${taskIdentifier}
```

dwelman: That was the cause of the problem in the first place, as it causes a conditional create failure

reinkrul (approved):

PR #340, opened 2025-01-28
Add ServiceBus demo client

dwelman:
In the move from kafka to service bus implementation the demo client was inadvertently removed, this change adds it back with accompanying tests


orchestrator/careplancontributor/ehr/servicebus_client.go
```diff
@@ -42,7 +52,15 @@ type NoopClient struct {
 // Returns an error if configuration is invalid or initialization fails.
 func NewClient(config ServiceBusConfig) (ServiceBusClient, error) {
 	ctx := context.Background()
+	log.Ctx(ctx).Info().Msgf("DAVIDTEST - %s", config.DemoEndpoint)
```
reinkrul, line 55: Cementing your legacy? :'-)

orchestrator/careplancontributor/ehr/servicebus_client.go
```diff
@@ -116,6 +134,31 @@ func (k *DebugClient) SubmitMessage(ctx context.Context, key string, value strin
 	return nil
 }
 
+// SubmitMessage submits a message to the configured endpoint over http.
+func (k *DemoClient) SubmitMessage(ctx context.Context, key string, value string) error {
+	jsonValue := strings.ReplaceAll(value, " ", "")
+	jsonValue = strings.ReplaceAll(jsonValue, "\n", "")
+	jsonValue = strings.ReplaceAll(jsonValue, "\t", "")
```
reinkrul, line 141: I don't think this is very smart, since it also removes spacing within strings. If you really want to do this, better unmarshal it then marshal again

reinkrul (changes requested):

orchestrator/careplancontributor/ehr/servicebus_client.go
dwelman: Hahaha oops

orchestrator/careplancontributor/ehr/servicebus_client.go
dwelman: That sounds like a better call, will do

reinkrul (approved):

PR #341, opened 2025-01-28
INT-514 | Read viewer onboarding tasks from delivered bundle

dwelman:
Copy functionality of BGZ page to the onboarding tasks as well, to read Tasks from the delivered bundle from the EHR if present


viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
```diff
@@ -12,34 +13,45 @@ export default async function AcceptedTaskOverview() {
...
+            console.log(`Found [${entries?.length}] Task resources`);
+        }
 
-        if (entry?.length) {
-            rows = entry.map((entry: any) => {
+        if (entries?.length) {
+            rows = entries.filter((entries) => entries.resource?.resourceType === "Task")
+                .map((entry: any) => {
                 const task = entry.resource as Task;
                 const bsn = task.for?.identifier?.value || "Unknown";
```
reinkrul, line 56: indenting looks off

reinkrul (approved):

viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
```diff
@@ -12,34 +13,45 @@ export default async function AcceptedTaskOverview() {
...
+            if (!response.ok) {
+                const errorText = await response.text();
+                console.error('Failed to fetch tasks: ', errorText);
+                throw new Error('Failed to fetch tasks: ' + errorText);
+            }
+
+            const responseBundle = await response.json() as Bundle;
+            entries = responseBundle.entry || [];
+            console.log(`Found [${entries?.length}] Task resources`);
+        }
```
reinkrul, line 50: I don't think we should fallback to the old behavior, since we want to get rid of it. If you agree, can you remove the accompanying code?

reinkrul (changes requested):

viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
dwelman: Sounds good, will do

viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
dwelman: Fixed

viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
dwelman: We can't remove this just yet as we need this to read tasks from fauxcare. I have added TODOs detailing that the logic needs to be removed in the future. When we have a better understanding of how we need to replace this logic I will create a ticket on the backlog

reinkrul: Note that Faux Care just sends it to Zorg bij jou the same way as Zorgplatform or any other hospital, so it shouldn't matter.

reinkrul (approved):

viewer_simulator/app/(DashboardLayout)/components/onboarding/accepted-task-overview.tsx
dwelman: Confirmed the bundle is being sent, seems there is a bug somewhere. Will find and fix

PR #352, opened 2025-01-29
Return notifications properly in create task handler

dmeijboom:
While looking at the code to see how a `CarePlan` is being created I noticed that the result handler of `handleCreateTask(..)` populates an array of notifications but incorrectly returns a hardcoded slice instead which results in a missing notification for the `CareTeam` in a task.

It seems like this should work similarly to `handleUpdateTask(..)` (which does return the `notifications` variable).


reinkrul (approved):

reinkrul:
I wrote a unit test, but I can't alter the PR. Feel free to merge, I'll make a PR for it.

PR #358, opened 2025-01-31
INT-568: remove OAuth2 Protected Resource Metadata for determining Authorization Server URL

reinkrul:
See related issue.

This PR removes the reliance on OAuth2 Protected Resource Metadata to resolve the Authorization Server URL, because:
- not all implementors support this
- in many cases, we know the identity of the party at which we want to get/send data, so we can just look up the Authorization Server URL in the CSD.

Test:
- [x] test environment
- [x] e2e test
- [x] local dev


orchestrator/careplanservice/service.go
```diff
@@ -144,9 +141,25 @@ func (s *Service) RegisterHandlers(mux *http.ServeMux) {
 	s.proxy = coolfhir.NewProxy("CPS->FHIR proxy", s.fhirURL, basePath,
 		s.orcaPublicURL.JoinPath(basePath), s.transport, true)
 	baseUrl := s.baseUrl()
-	s.profile.RegisterHTTPHandlers(basePath, baseUrl, mux)
 
 	// Binding to actual routing
+	// Metadata
+	mux.HandleFunc("GET "+basePath+"/metadata", func(httpResponse http.ResponseWriter, request *http.Request) {
```
reinkrul, line 147: @bramwesselo does this look right?

dwelman (approved):
I am not an auth expert so can't comment on the finer points. Code looks sound and you have done extensive testing so I am happy with it

PR #360, opened 2025-02-02
INT-569: Support DELETE when unmanaged operations are allowed

reinkrul:
Useful for cleaning up test data, since testing sometimes involves creating a new patient in the downstream EHR.

This allows parties to delete the related resources (e.g. Task, CarePlan, ServiceRequest and Patient) and re-send the order.

A lot of the changes are associated with FHIR API returning no response body in case the delete succeeds (204 No Content)

### DIFF
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -150,7 +150,7 @@ func (s *Service) RegisterHandlers(mux *http.ServeMux) {
 	// Creating a resource
 	mux.HandleFunc("POST "+basePath+"/{type}", s.profile.Authenticator(baseUrl, func(httpResponse http.ResponseWriter, request *http.Request) {
 		resourceType := request.PathValue("type")
-		s.handleCreateOrUpdate(request, httpResponse, resourceType, "CarePlanService/Create"+resourceType)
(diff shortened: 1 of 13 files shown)


orchestrator/careplanservice/service.go
```diff
@@ -266,10 +277,14 @@ func (s *Service) handleUnmanagedOperation(request FHIRHandlerRequest, tx *coolf
 
 func (s *Service) handleCreateOrUpdate(httpRequest *http.Request, httpResponse http.ResponseWriter, resourcePath string, operationName string) {
```
dwelman, line 278: Not a blocker as I know this work is needed for testing tomorrow, can be changed in the future:

We are now using this method for CREATE, PUT, and DELETE. I was initially confused why it was being used for delete, maybe it should be renamed for clarity?

dwelman (approved):

orchestrator/careplanservice/service.go
reinkrul: Suggestions for naming?

dwelman: Hmm, I don't have any great ones. Maybe something like `handleResourceModification`?

reinkrul: I renamed it to `handleModification`, for parity with `handleGet` and `handleSearch`

PR #379, opened 2025-02-05
INT-589 | Always send bundle to demo endpoint if set

dwelman:
Creates a demo messenger client if the demo viewer endpoint is set, and the app is not running in strict mode. This way, bundles are delivered to the configured viewer over HTTP for demo purposes, and the existing serrvicebus client still runs, allowing us to test both

### DIFF
--- a/orchestrator/careplancontributor/ehr/servicebus_client.go
+++ b/orchestrator/careplancontributor/ehr/servicebus_client.go
@@ -33,7 +33,8 @@ type ServiceBusClient interface {
 
 // ServiceBusClientImpl provides an implementation for interacting with Azure Service Bus using the provided configuration.
 type ServiceBusClientImpl struct {
-	config ServiceBusConfig
+	config     ServiceBusConfig
+	demoClient *DemoClient
 }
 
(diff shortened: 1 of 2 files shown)


orchestrator/careplancontributor/ehr/servicebus_client.go
```diff
@@ -118,6 +121,15 @@ func (k *ServiceBusClientImpl) SubmitMessage(ctx context.Context, key string, va
 		return err
 	}
 	log.Ctx(ctx).Debug().Msgf("SubmitMessage, submitted key %s", key)
+
+	if k.demoClient != nil {
+		err = k.demoClient.SubmitMessage(ctx, key, value)
+		if err != nil {
+			log.Ctx(ctx).Error().Err(err).Msgf("DemoClient submission failed with %s", err.Error())
+			return err
```
dwelman, line 129: I am not sure about returning an error here, as we may just want to log the error and keep going. That being said, this is at the end of the method so it probably isn't an issue

reinkrul (approved):

PR #385, opened 2025-02-07
Remove fallback in viewer that fetches when task/careplan is not set …

dwelman:
…locally


dwelman:
I'll implement this in a different PR

PR #386, opened 2025-02-07
INT-592: Event handling setup

reinkrul:
This is the initial PR that adds an event handling layer. The plan is to convert all pieces of business logic that is triggered by FHIR notifications to this model:

- When a Task is accepted, a Bundle with all enrolment info is created and send over Azure ServiceBus.
- When a new Task is created, the Task Engine will negotiate it with the Task Filler
- When a CarePlan is created, it is registered on the NVI (Nationale VerwijsIndex). It will be a simple webhook event handler.


dwelman (approved):

dwelman:
Looks good! Not questions we need to have an answer for now but just to keep in mind as we go forward:

* How will this interact with the servicebus impl. in CPC, if at all? 
* Do we have an idea for how we want to handle failed event retry?

dwelman:
Sounds good!

PR #391, opened 2025-02-10
Temporarily disable Servicebus implementation

dwelman:
Bypass servicebus so that only demo client runs, this is a stopgap measure while there are problems with servicebus in our azure environment, and should be reverted ASAP


reinkrul:
Why not send it over the demo client first, then to ServiceBus? Do we then still need to disable ServiceBus?

dwelman:
Honestly... Much better idea, let me do that instead. Then it doesn't matter if ServiceBus fails or not

PR #395, opened 2025-02-11
INT-386 | Disable requirement for CPS URL in CPC

dwelman:
CPS is not required in CPC as we add support for multiple CPS's, if the CPS URL is not set then the proxy method will fail


dwelman:
I missed one check in the config, removing that then I will update the PR

reinkrul (approved):

PR #396, opened 2025-02-11
INT-596 | Implement webhook handler

dmeijboom:
Basic webhook handler implementation. Sends a POST request containing the full resource encoded in JSON. Expects a HTTP 200 to succeed. Also tested manually.

### DIFF
--- a/orchestrator/careplancontributor/webhook/eventhandler.go
+++ b/orchestrator/careplancontributor/webhook/eventhandler.go
@@ -1,16 +1,62 @@
 package webhook
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
 	events "github.com/SanteonNL/orca/orchestrator/careplancontributor/event"
+	"github.com/rs/zerolog/log"
 )
+
+// See: https://go.googlesource.com/go/+/refs/tags/go1.20.4/src/net/http/server.go#1098
+const maxPostHandlerReadBytes = 256 << 10
 
 var _ events.Handler = EventHandler{}
 
+func disposeResponseBody(r io.ReadCloser) {
+	if _, err := io.CopyN(io.Discard, r, maxPostHandlerReadBytes); err != nil {
+		log.Error().Err(err).Msg("failed to read request body")
+	}
+}
+
 type EventHandler struct {
-	URL string
+	client *http.Client
+	URL    string
+}
+
+func NewEventHandler(url string) EventHandler {
+	return EventHandler{URL: url, client: &http.Client{}}
 }
 
 func (w EventHandler) Handle(ctx context.Context, event events.Instance) error {
-	panic("implement me")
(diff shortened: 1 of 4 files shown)


orchestrator/careplancontributor/webhook/eventhandler.go
```diff
@@ -1,16 +1,50 @@
...
 
 func (w EventHandler) Handle(ctx context.Context, event events.Instance) error {
-	panic("implement me")
+	data, err := json.Marshal(event.FHIRResource)
+	if err != nil {
+		return fmt.Errorf("failed to serialize event: %w", err)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, "POST", w.URL, bytes.NewReader(data))
+	if err != nil {
```
reinkrul, line 32: at least add content-type

orchestrator/careplancontributor/webhook/eventhandler.go
```diff
@@ -1,16 +1,50 @@
...
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+
+	response, err := w.client.Do(request)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+
+	defer response.Body.Close()
```
reinkrul, line 41: we might need to drain the response body, not sure. Was previously required, but maybe not any more.

orchestrator/careplancontributor/webhook/eventhandler_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"testing"
+
+	events "github.com/SanteonNL/orca/orchestrator/careplancontributor/event"
+	"github.com/stretchr/testify/require"
+)
+
+type unmarshallable struct{}
+
+func (u unmarshallable) MarshalJSON() ([]byte, error) {
+	return nil, errors.New("fail")
```
reinkrul, line 18: I don't really care about these test cases (Marshal only fals if it's passed a function pointer), but it doesn't hurt much.

orchestrator/careplancontributor/webhook/eventhandler_test.go
```diff
@@ -0,0 +1,102 @@
+package webhook
```
reinkrul, line 1: good tests, just a bit overcomplete i.m.o.

reinkrul (approved):
small thing left to fix

orchestrator/careplancontributor/webhook/eventhandler.go
dmeijboom: After reading https://github.com/golang/go/issues/60240 I'm still not sure, let's do it anyways

orchestrator/careplancontributor/webhook/eventhandler_test.go
dmeijboom: I tend to test all branches but it might be a bit much

PR #416, opened 2025-02-16
Set careplan author on creation

dmeijboom:
According to the [SCP specification](https://santeonnl.github.io/shared-care-planning/overview.html#careplan) the author should be set for a CarePlan. However, whenever a new CarePlan is created (based on a Task), it doesn't set the author.


reinkrul (changes requested):
Can you add an assertion to the unit test for this?

dmeijboom:
> Can you add an assertion to the unit test for this?

Yes, definitely. There are no assertions on the CarePlan yet so I'll add all of them

reinkrul (approved):
cheers!

PR #421, opened 2025-02-20
INT-621 | Verify notification bundle referential integrity

dwelman:
When a Search operation is performed on a FHIR resource and that resource is not found, no error is returned. If a resource we expect to be in the notification bundle is not found, it will throw an error


orchestrator/careplancontributor/ehr/notification_bundle.go
```diff
@@ -0,0 +1,271 @@
...
+		if refType == "CarePlan" {
+			values.Set("_include", "CarePlan:care-team")
+		}
+		err := cpsClient.SearchWithContext(ctx, refType, values, &bundle)
+		if err != nil {
+			return nil, err
+		}
+		bundles = append(bundles, bundle)
+
+		if len(refIds) != len(bundle.Entry) {
```
dwelman, line 175: NB: Most of the changes in this PR are just moving the code to its own file so it is easier to read and test and for the future where we may want to notify on more than just task. This if statement is the actual check

reinkrul (approved):

PR #422, opened 2025-02-20
INT-626: Hospital Simulator | BUG, mock BSN is not 9 digits

rolandgroen:
The BSN is defined as a 9 digit number. Using something different breaks stuff on the other side of the integration.

Fix: Limit patient BSN to 9 characters in service request. Updated the logic for generating patient BSN to ensure it is a 9-character string.

### DIFF
--- a/hospital_simulator/app/(DashboardLayout)/components/service-requests/create-service-request-dialog.tsx
+++ b/hospital_simulator/app/(DashboardLayout)/components/service-requests/create-service-request-dialog.tsx
@@ -174,7 +174,8 @@ function createServiceRequestBundle(firstName: string, lastName: string, conditi
     }
     const performerIdentifier = performer.identifier[0]
 
-    const patientBsn = Date.now();
+    let patientBsn = "" + Date.now()
+    patientBsn = patientBsn.substring(patientBsn.length - 9, patientBsn.length)
 
     return {
         "resourceType": "Bundle",


hospital_simulator/app/(DashboardLayout)/components/service-requests/create-service-request-dialog.tsx
```diff
@@ -174,7 +174,8 @@ function createServiceRequestBundle(firstName: string, lastName: string, conditi
     }
     const performerIdentifier = performer.identifier[0]
 
-    const patientBsn = Date.now();
+    let patientBsn = "" + Date.now()
+    patientBsn = patientBsn.substring(patientBsn.length - 9, patientBsn.length)
```
JorisHeadease, line 178: This code doesn't create "valid" BSNs (as those must comply with the "elfproef"). If they only care about the length of the BSN, then the solution seems fine. 

Maybe check with Dirk if they need that?

reinkrul: Maybe we should take the extra step to generate actually valid test BSNs? They are generally in the 9999 range (https://www.rvig.nl/sites/default/files/2023-05/Test%20burgerservicenummers%20en%20A-nummers%20inclusief%20omnummertabel.ods)?

Then they are never confused as being actual "real" BSNs

rolandgroen: Seems like the best option to me.

reinkrul (approved):

hospital_simulator/app/(DashboardLayout)/components/service-requests/create-service-request-dialog.tsx
reinkrul: Let's do that in a follow-up PR

PR #425, opened 2025-02-20
INT-573: Added a styled error page for uncaught exceptions. 

JorisHeadease:
This error page contains a retry button and reports the `digest` and UTC timestamp of the error. The page instructs the user to send this information to support if the issue persists. The digest can be used to correlate to the error in the logs.

``` GET /enrollment/new 200 in 116ms
 ⨯ Error: Not implemented
    at ConfirmDataPreEnrollment (app/enrollment/new/page.tsx:8:8)
   6 | export default function ConfirmDataPreEnrollment() {
   7 |
>  8 |   throw new Error('Not implemented')
     |        ^
   9 |
  10 |   return (
  11 |     <Card className="border-0 shadow-none px-0"> {
  digest: '4196643865'
```

### Additional configuration
Added the `SUPPORT_CONTACT_LINK` env var to the frontend. When this is set, the "Neem contact op" subtext appears at the bottom, the the value is set as the anchor href. This can be a website, but also a `mailto:` of `tel:` link etc.

## Style
Screenshot 2025-02-20 at 20 38 30

### DIFF
--- a/frontend/.env.local.example
+++ b/frontend/.env.local.example
@@ -1,1 +1,2 @@
 TERMINOLOGY_SERVER_BASE_URL=https://terminologieserver.nl/fhir
+SUPPORT_CONTACT_LINK=mailto:support@fauxcare.zorgbijjou.com #This link will be shown at the bottom of the error.tsx page if configured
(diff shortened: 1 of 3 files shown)


reinkrul (commented):
Very nice. Where does "digest" come from, and how to relate it back to the application logging?

dwelman (approved):

PR #428, opened 2025-02-21
INT-572: Added 3 retries in the ZPF handleLaunch when the `getSession…

JorisHeadease:
…Data() returns an error. During testing moments with ChipSoft, we seemed to sometiems get a 404 when a newly created workflow was quickly launched and the code tried to query the `Patient` bound to that workflow.

### DIFF
--- a/orchestrator/careplancontributor/applaunch/zorgplatform/service_test.go
+++ b/orchestrator/careplancontributor/applaunch/zorgplatform/service_test.go
@@ -276,6 +276,13 @@ func TestService(t *testing.T) {
 	t.Run("retry getSessionData 3 times - should work", func(t *testing.T) {
 		globals.CarePlanServiceFhirClient = &test.StubFHIRClient{}
 
+		// Save the original sleep function and restore it after the test.
+		originalSleep := sleep
+
+		// Override sleep to collect the duration of the sleep requested duration instead of actually sleeping.
+		var sleeps []time.Duration
+		sleep = func(t time.Duration) { sleeps = append(sleeps, t) }
+
 		// Mock getSessionData to fail twice before succeeding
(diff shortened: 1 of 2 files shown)


orchestrator/careplancontributor/applaunch/zorgplatform/service_test.go
```diff
@@ -271,6 +272,46 @@ func TestService(t *testing.T) {
...
 	})
+
+	t.Run("retry getSessionData 3 times - should work", func(t *testing.T) {
+		globals.CarePlanServiceFhirClient = &test.StubFHIRClient{}
+
+		// Mock getSessionData to fail twice before succeeding
+		callCount := 0
+		service.getSessionData = func(ctx context.Context, accessToken string, launchContext LaunchContext) (*user.SessionData, error) {
+			callCount++
+			if callCount < 3 {
```
dwelman, line 283: I'm a bit hesitant putting a unit test in with a 1200ms delay as that is going to slow us down every time we run our tests. Could you make the 200ms sleep value configurable in the service/client and then at least it can be set to something more sensible for unit tests?

reinkrul: or make `sleep()` mockable;

```golang
var sleep = time.Sleep

// in test:
var sleeps []time.Duration 
sleep = func(t time.Duration) { sleeps = append(sleeps, sleep)}
```

(don't forget to restore in `Cleanup()`)

JorisHeadease: Cheers, hopefully solved this in the latest commit!

dwelman (approved):

PR #438, opened 2025-02-25
INT-630 | Contain careteam within a careplan

dmeijboom:
This PR refactors ORCA to "contain" a 'CareTeam' within a 'CarePlan'. It works under the assumption that each 'CarePlan' has only a single 'CareTeam', which should be acceptable given that multiple unit tests maintain this same assumption.

There have been numerous changes, and I've attempted to be careful while implementing them. However, we need to verify that all business rules continue to function as they did previously.


orchestrator/careplancontributor/service.go
```diff
@@ -438,9 +438,9 @@ func (s Service) authorizeScpMember(request *http.Request) (*ScpValidationResult
 		return nil, err
 	}
 
-	// Use extract CarePlan ID to be used for our query that will get the CarePlan and CareTeam in a bundle
+	// Use extract CarePlan ID to be used for our query that will get the CarePlan in a bundle
 	var bundle fhir.Bundle
-	err = cpsFHIRClient.Search("CarePlan", url.Values{"_id": {carePlanId}, "_include": {"CarePlan:care-team"}}, &bundle)
+	err = cpsFHIRClient.Search("CarePlan", url.Values{"_id": {carePlanId}}, &bundle)
```
reinkrul, line 443: It's not wrong, but we can now replace this with a simple `Read()` (or better, `ReadWithContext`). Then you don't need to deal with the bundle searchresult

orchestrator/careplancontributor/service.go
```diff
@@ -487,7 +482,7 @@ func (s Service) authorizeScpMember(request *http.Request) (*ScpValidationResult
 	}
 	return &ScpValidationResult{
 		carePlan:  &carePlan,
-		careTeams: &careTeams,
+		careTeams: &[]fhir.CareTeam{*careTeam},
```
reinkrul, line 485: @bramwesselo do we need to support multiple CareTeams, or can we just settle on "1 CarePlan has exactly 1 (contained) CareTeam"?

orchestrator/lib/coolfhir/contained.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+var (
+	ErrCareTeamNotFound = errors.New("CareTeam not found in FHIR Bundle")
```
reinkrul, line 12: I think they'r not in bundles anymore now? (but contained)

orchestrator/lib/coolfhir/contained.go
```diff
@@ -0,0 +1,98 @@
...
+	r.Raw = data
+
+	return nil
+}
+
+func (r *ContainedResource) MarshalJSON() ([]byte, error) {
+	return r.Raw, nil
+}
+
+func UpdateContainedResource(contained json.RawMessage, id *fhir.Reference, resourceType string, resource any) (json.RawMessage, error) {
```
reinkrul, line 35: a `fhir.Reference` already contains a type (although it might be nil). So just use the reference and remove the `id` parameter, or change the type of `id` to `string`.

orchestrator/lib/coolfhir/contained.go
```diff
@@ -0,0 +1,98 @@
...
+	var resources []ContainedResource
+
+	if err := json.Unmarshal(contained, &resources); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal contained resources: %w", err)
+	}
+
+	ref := *id.Reference
+
+	for i, containedResource := range resources {
+		if containedResource.ID == ref[1:] && containedResource.Type == resourceType {
```
reinkrul, line 54: Tricky, do contained resources always follow the `#...` format? Maybe at least add a comment

I think we only actually support 1 CareTeam? Then you could also just find the contained resource to update by resource type.

orchestrator/lib/coolfhir/contained.go
reinkrul, line 35: `resource`  is a slice of contained resources, so maybe just call it `containedResouces`?

orchestrator/lib/coolfhir/contained.go
```diff
@@ -0,0 +1,98 @@
...
+			resources[i] = containedResource
+
+			return json.Marshal(resources)
+		}
+	}
+
+	return nil, fmt.Errorf("contained resource not found")
+}
+
+func ResolveCareTeam(carePlan *fhir.CarePlan) (*fhir.CareTeam, error) {
```
reinkrul, line 65: `CareTeamFromCarePlan` or make a more generic function `GetContainedResource` or something, in which you pass the struct to unmarshal into.
Something like we do in `bundle.go`:

`func GetContainedResource(container any, target any, predicate func()) error` (unmarshal into target)

orchestrator/lib/coolfhir/util.go
```diff
@@ -188,19 +188,21 @@ func parseTimestamp(timestampString string) (time.Time, error) {
...
+		for _, identifier := range principalOrganizationIdentifiers {
+			if IdentifierEquals(participant.Member.Identifier, &identifier) {
+				return &participant
 			}
 		}
 	}
 	return nil
 }
 
+func IsLocalReference(reference *fhir.Reference) bool {
```
reinkrul, line 202: The term is "local relative reference" (local reference can also be `<type>/<id>` I think)

orchestrator/careplanservice/subscriptions/manager.go
```diff
@@ -75,22 +75,15 @@ func (r DerivingManager) Notify(ctx context.Context, resource interface{}) error
...
-
-			var careTeam fhir.CareTeam
-
-			if err := r.FhirClient.ReadWithContext(ctx, *team.Reference, &careTeam); err != nil {
-				log.Ctx(ctx).Error().Msgf("failed to read CareTeam in FHIR store %s: %s", *team.Reference, err)
-				continue
-			}
+		careTeam, err := coolfhir.ResolveCareTeam(carePlan)
+		if err != nil {
+			log.Ctx(ctx).Error().Msgf("failed to read CareTeam in CarePlan %s", err)
```
reinkrul, line 80: have the error populate the structurized logging field meant for it:
```suggestion
			log.Ctx(ctx).Err(err).Msg("failed to read CareTeam in CarePlan")
```

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -61,35 +52,24 @@ func (s *Service) handleSearchCarePlan(ctx context.Context, queryParams url.Valu
...
+		err = validatePrincipalInCareTeams(principal, []fhir.CareTeam{*ct})
+		if err != nil {
+			continue
 		}
+
+		filterRefs = append(filterRefs, "CarePlan/"+*cp.Id)
 	}
-	retBundle := filterMatchingResourcesInBundle(ctx, bundle, []string{"CarePlan", "CareTeam"}, filterRefs)
+
+	retBundle := filterMatchingResourcesInBundle(ctx, bundle, []string{"CarePlan"}, filterRefs)
```
reinkrul, line 72: I don't think we need this call, since this is to filter resources that were returned (due to the CareTeam include, previously) but weren't requested by the client. Right @dwelman ?

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -227,15 +234,19 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 					Type:      to.Ptr("Task"),
 				},
 			})
-			tx.Update(carePlan, "CarePlan/"+*carePlan.Id)
+
+			updated, err := careteamservice.Update(s.fhirClient, *carePlan.Id, task, tx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to update CarePlan: %w", err)
```
reinkrul, line 240:
```suggestion
				return nil, fmt.Errorf("failed to update CareTeam: %w", err)
```

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -227,15 +234,19 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
+			}
+
+			if !updated {
```
reinkrul, line 243: this looks like the the wrong negation? No tests failing on this?

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -146,33 +159,25 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
-			Identifier: &principal.Organization.Identifier[0],
-			Type:       to.Ptr("Organization"),
-		})
-		if !ok {
-			return nil, errors.New("failed to activate membership for new CareTeam")
-		}
-
 		tx.Create(carePlan, coolfhir.WithFullUrl(carePlanURL)).
-			Create(careTeam, coolfhir.WithFullUrl(careTeamURL)).
+			Create(careTeam, coolfhir.WithFullUrl(careTeamID)).
```
reinkrul, line 163: if it's a contained resource, don't create it?

reinkrul (changes requested):

reinkrul:
@bramwesselo this PR removes support for searching and reading CareTeams, but SCP still specifies it: https://santeonnl.github.io/shared-care-planning/security-authorization.html#resource-access

So either we have to keep it in ORCA, or remove it from SCP

reinkrul:
Good simplification!

orchestrator/careplancontributor/service.go
bramwesselo: we can settle on: 1 CarePlan has exactly 1 (contained) CareTeam

orchestrator/lib/coolfhir/contained.go
dwelman: Yeah we have `CarePlan.CareTeam`

`CareTeam              []Reference        `bson:"careTeam,omitempty" json:"careTeam,omitempty"``

orchestrator/careplanservice/handle_getcareplan.go
dwelman: Correct. We agreed that it is acceptable to return the CareTeam alongside the CarePlan. If CareTeam is now contained within CarePlan then it shouldn't be needed at all.

orchestrator/careplancontributor/service.go
dwelman: This would also be my preference, less code, less overhead, more performant

orchestrator/careplanservice/handle_getcareplan.go
dwelman: It looks like this is still being used to filter out resources the requester does not have access to, so I think it can stay as is. 

`filterRefs = append(filterRefs, "CarePlan/"+*cp.Id)`

orchestrator/careplanservice/util.go
```diff
@@ -19,12 +20,11 @@ import (
 func GetCarePlanAndCareTeams(ctx context.Context, client fhirclient.Client, carePlanReference string) (fhir.CarePlan, []fhir.CareTeam, *fhirclient.Headers, error) {
```
dwelman, line 20: The method was needed when we had to search for a CarePlan and include CareTeam separately, this can be replaced by simply fetching the CarePlan and using the contained CareTeam now

orchestrator/lib/coolfhir/contained.go
dmeijboom: I'm not sure if I understand what you mean. `resource` (in the arguments) is a single resource (this could be a `CareTeam` for example)

orchestrator/careplanservice/handle_createtask.go
dmeijboom: This looks weird but is actually correct. The `careteamservice` already adds the CarePlan to the transaction but only if it was updated. If it's not updated we still want to update the careplan (since the activity changed)

orchestrator/careplancontributor/service.go
```diff
@@ -331,24 +329,30 @@ func (s *Service) proxyToAllCareTeamMembers(writer http.ResponseWriter, request
...
+
+	err = cpsFHIRClient.ReadWithContext(request.Context(), carePlanRef, &carePlan)
+	if err != nil {
+		return fmt.Errorf("failed to get CarePlan %s for X-SCP-Context: %w", carePlanRef, err)
+	}
+
+	careTeam, err := coolfhir.CareTeamFromCarePlan(&carePlan)
 	if err != nil {
-		return fmt.Errorf("failed to get CarePlan and CareTeams for X-SCP-Context: %w", err)
+		return fmt.Errorf("failed to resolve CareTeam %s: %w", carePlanRef, err)
```
reinkrul, line 342:
```suggestion
		return fmt.Errorf("failed to resolve CareTeam (carePlan=%s): %w", carePlanRef, err)
```

orchestrator/careplancontributor/service.go
```diff
@@ -438,31 +442,22 @@ func (s Service) authorizeScpMember(request *http.Request) (*ScpValidationResult
 		return nil, err
 	}
 
-	// Use extract CarePlan ID to be used for our query that will get the CarePlan and CareTeam in a bundle
-	var bundle fhir.Bundle
-	err = cpsFHIRClient.Search("CarePlan", url.Values{"_id": {carePlanId}, "_include": {"CarePlan:care-team"}}, &bundle)
+	var carePlan fhir.CarePlan
+
+	err = cpsFHIRClient.ReadWithContext(request.Context(), "CarePlan/"+carePlanId, &carePlan)
```
reinkrul, line 447:
```suggestion
	err = cpsFHIRClient.ReadWithContext(request.Context(), carePlanRef, &carePlan)
```

The ID parsing can then be removed, probably

orchestrator/careplancontributor/service.go
```diff
@@ -438,31 +442,22 @@ func (s Service) authorizeScpMember(request *http.Request) (*ScpValidationResult
...
-		return nil, coolfhir.NewErrorWithCode("CareTeam not found in bundle", http.StatusNotFound)
-	}
-
-	var carePlan fhir.CarePlan
-	err = coolfhir.ResourceInBundle(&bundle, coolfhir.EntryIsOfType("CarePlan"), &carePlan)
 
+	careTeam, err := coolfhir.CareTeamFromCarePlan(&carePlan)
 	if err != nil {
-		return nil, err
+		return nil, coolfhir.NewErrorWithCode("unable to resolve CareTeam", http.StatusNotFound)
```
reinkrul, line 460: Hmm, maybe InternalServerError is better? Since there shouldn't be CarePlans without CareTeam. 4xx indicates the client can do something to fix it

reinkrul (changes requested):

reinkrul:
Small things

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -177,27 +165,43 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 			Type:       to.Ptr("Organization"),
 		})
 
+		data, err := json.Marshal([]any{careTeam})
+		if err != nil {
+			return nil, coolfhir.NewErrorWithCode("failed to marshal CareTeam", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
dmeijboom: Fair. This is virtually impossible but yeah

dwelman (approved):

reinkrul (approved):
Let's see it in action!

PR #440, opened 2025-02-26
INT-337 | Create audit events in CPS

dwelman:
Audit events are created for CRUD methods, these will be used in the future to validate resource access as well as for compliance on resource access

### DIFF
--- a/orchestrator/careplanservice/handle_createtask.go
+++ b/orchestrator/careplanservice/handle_createtask.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/SanteonNL/orca/orchestrator/lib/audit"
+
 	fhirclient "github.com/SanteonNL/go-fhir-client"
 	"github.com/SanteonNL/orca/orchestrator/lib/auth"
 	"github.com/google/uuid"
@@ -154,11 +156,47 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 			return nil, errors.New("failed to activate membership for new CareTeam")
 		}
 
+		carePlanAuditEvent, err := audit.AuditEvent(ctx, fhir.AuditEventActionC, &fhir.Reference{
+			Reference: to.Ptr(carePlanURL),
+			Type:      to.Ptr("CarePlan"),
+		}, &fhir.Reference{
+			Identifier: task.Requester.Identifier,
+			Type:       to.Ptr("Organization"),
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		careTeamAuditEvent, err := audit.AuditEvent(ctx, fhir.AuditEventActionC, &fhir.Reference{
+			Reference: to.Ptr(careTeamURL),
+			Type:      to.Ptr("CareTeam"),
+		}, &fhir.Reference{
+			Identifier: task.Requester.Identifier,
+			Type:       to.Ptr("Organization"),
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		taskAuditEvent, err := audit.AuditEvent(ctx, fhir.AuditEventActionC, &fhir.Reference{
+			Reference: to.Ptr(taskURL),
+			Type:      to.Ptr("Task"),
+		}, &fhir.Reference{
+			Identifier: task.Requester.Identifier,
+			Type:       to.Ptr("Organization"),
+		})
+		if err != nil {
+			return nil, err
+		}
+
 		tx.Create(carePlan, coolfhir.WithFullUrl(carePlanURL)).
 			Create(careTeam, coolfhir.WithFullUrl(careTeamURL)).
-			Create(task, coolfhir.WithFullUrl(taskURL), coolfhir.WithRequestHeaders(request.HttpHeaders))
-		careTeamEntryIdx = len(tx.Entry) - 2
-		taskEntryIdx = len(tx.Entry) - 1
+			Create(task, coolfhir.WithFullUrl(taskURL), coolfhir.WithRequestHeaders(request.HttpHeaders)).
+			Create(carePlanAuditEvent).
+			Create(careTeamAuditEvent).
+			Create(taskAuditEvent)
+		careTeamEntryIdx = len(tx.Entry) - 5
+		taskEntryIdx = len(tx.Entry) - 4
 		taskBundleEntry = tx.Entry[taskEntryIdx]
 		carePlanBundleEntry = &tx.Entry[0]
 		carePlanBundleEntryIdx = 0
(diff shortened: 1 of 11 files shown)


orchestrator/lib/audit/audit.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		Entity: []fhir.AuditEventEntity{
+			{
+				What: resourceReference,
+			},
+		},
+		Recorded: TimeNow().Format(time.RFC3339),
+		Action:   to.Ptr(fhir.AuditEventAction(action)),
+		Source: fhir.AuditEventSource{
+			// Local organisation i.e. orchestrator
```
dwelman, line 37: We use the same org identifier to determine if a node can create a CarePlan, is this the correct org to use?

orchestrator/careplanservice/service.go
```diff
@@ -388,13 +395,38 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
 		coolfhir.WriteOperationOutcomeFromError(httpRequest.Context(), err, operationName, httpResponse)
 		return
 	}
+
+	auditEvent, err := audit.AuditEvent(httpRequest.Context(),
+		fhir.AuditEventActionR, &fhir.Reference{
+			Reference: to.Ptr(resourceType + "/" + resourceId),
+			Type:      to.Ptr(resourceType),
+		}, &fhir.Reference{
+			Identifier: &principal.Organization.Identifier[0],
```
dwelman, line 404: Is this correct? this would essentially mean the agent actioning and observing the audit is the same

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -154,11 +156,47 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
...
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		careTeamAuditEvent, err := audit.AuditEvent(ctx, fhir.AuditEventActionC, &fhir.Reference{
+			Reference: to.Ptr(careTeamURL),
+			Type:      to.Ptr("CareTeam"),
+		}, &fhir.Reference{
+			Identifier: task.Requester.Identifier,
```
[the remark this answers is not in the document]
reinkrul: Hm, you could argue that is was the CarePlanService that actioned the audit event. But it doesn't really matter too much, since when Dillen's PR is merged this audit event will be removed anyways?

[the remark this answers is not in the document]
reinkrul: So maybe just already remove it?

orchestrator/lib/audit/audit.go
reinkrul: Not sure what the purpose is, but now the `AuditEvent.Source.Observer` is always the same as zAuditEvent.Agent`? Because you generally pass the `Task.requester` or principal.organization.identifier for `actingAgentRef`, and then get the authenticated principal (is caller identity) from the context, which are the same.

Shouldn't `Observer` be populated with the identity of the local node? Thus fetching the local identity (can be multiple in theory, but only 1 in practice) and taking that as observer?

What does @bramwesselo think?

reinkrul: So disctintion is, principal = authenticated client (remote party), local identity = the local ORCA instance

orchestrator/careplanservice/service.go
```diff
@@ -388,13 +395,38 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
...
+			Type:       to.Ptr("Organization"),
+		})
+	if err != nil {
+		coolfhir.WriteOperationOutcomeFromError(httpRequest.Context(), err, operationName, httpResponse)
+		return
+	}
+	err = s.fhirClient.Create(auditEvent, &auditEvent)
```
reinkrul, line 411: with context

reinkrul (changes requested):
Pretty neat in general, some remarks

orchestrator/lib/audit/audit.go
dwelman: Yeah that sounds better but I will confer with Bram as well, I didn't know what to set it as but wanted to set it in the meantime so I don't forget to add it later and we already have code in place for it

orchestrator/lib/audit/audit.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
+	"github.com/SanteonNL/orca/orchestrator/lib/to"
+
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+var TimeNow = time.Now
+
+func AuditEvent(ctx context.Context, action fhir.AuditEventAction, resourceReference *fhir.Reference, actingAgentRef *fhir.Reference) (*fhir.AuditEvent, error) {
```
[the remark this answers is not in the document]
dwelman: Ah oops, it used to be in a different package before I moved it to use the auth package, I will fix this

orchestrator/lib/audit/audit.go
```diff
@@ -0,0 +1,68 @@
...
+	"testing"
+	"time"
+
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
+	"github.com/SanteonNL/orca/orchestrator/lib/to"
+
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+var TimeNow = time.Now
```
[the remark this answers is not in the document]
dwelman: Good point, will change

orchestrator/lib/audit/audit.go
bramwesselo: the `AuditEvent.Source` should be the 'local node'

orchestrator/careplanservice/config.go
```diff
@@ -23,5 +26,11 @@ func (c Config) Validate() error {
 	if c.FHIR.BaseURL == "" {
 		return errors.New("careplanservice.fhir.url is not configured")
 	}
+	if c.AuditObserverSystem == "" {
+		return errors.New("careplanservice.auditobserversystem is not configured")
+	}
+	if c.AuditObserverValue == "" {
+		return errors.New("careplanservice.auditobservervalue is not configured")
+	}
```
reinkrul, line 34: Do we want the URA number here? We can just defer them from the local node (`profile.Identity()`)

reinkrul, line 34: Do we want the URA number here? We can just defer them from the local node (`profile.Identity()`)

reinkrul: My thinking is, less configuration options = simpler system. Also, if we ever were to make ORCA multi-tenant, you'd want the tenant (thus, this identity) in the audit i.m.o.

orchestrator/lib/audit/audit.go
```diff
@@ -39,6 +29,10 @@ func Event(action fhir.AuditEventAction, resourceReference *fhir.Reference, acti
 		interactionDisplay = "Search"
 	}
 
+	if localIdentity == nil {
```
reinkrul, line 32: then just make it a non-pointer here, then you don't need to nil-check (`getLocalIdentity()` already returns an error, so it can do the nil check)

reinkrul (changes requested):

reinkrul (approved):

PR #454, opened 2025-02-27
INT-645 | Task.for must be the same patient as CarePlan.subject

dwelman:


orchestrator/careplanservice/handle_createtask.go
```diff
@@ -175,6 +175,11 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 			return nil, err
 		}
 
+		// Validate Task.for matches CarePlan.subject
+		if task.For == nil || !coolfhir.LogicalReferenceEquals(*task.For, carePlan.Subject) {
```
reinkrul, line 179: This now requires the patient ref as logical identifier for both CarePlan and Task. @bramwesselo we should then require this in the spec? Note that the CarePlan.subject is copied from the first primary Task that triggers the CarePlan creation.

@dwelman can you check whether the Zorgplatform app launch sets the logical identifier (BSN) on the patient ref?

dwelman: Sure, I'll run some tests in the morning

bramwesselo: I think a resolvable reference is a MUST, the identifier (BSN) is a SHOULD (just like we discussed/settled on lately): We wouldn't want to specify which identifiers should be added in references (messy/custom). If we'd be able to resolve the reference of both CarePlan.subject and Task.for, this check should be possible. If we don't have access to the patient in Task.for, we're in a situation were the CarePlan is just being created (and then they're the same anyway)

reinkrul (approved):

PR #458, opened 2025-03-03
INT-648: Add patient BSN to Tasks created through Frontend

reinkrul:


frontend/lib/fhirUtils.ts
```diff
@@ -194,9 +201,14 @@ export const constructTaskBundle = (serviceRequest: ServiceRequest, primaryCondi
     type: "transaction"
 } => {
     const cleanedPatient = cleanPatient(patient);
+    const patientReference : Reference = {
+        type: "Patient",
+        reference: "urn:uuid:patient",
+        identifier: patient.identifier?.filter((identifier) => identifier.system === BSN_SYSTEM)[0]!
```
dwelman, line 207: This could result in an unhandled error, could we wrap this in a try catch?

dwelman (approved):

frontend/lib/fhirUtils.ts
reinkrul: And it probably shouldn't be hardcoded support BSN only. So maybe we just take the first identifier or something... Or configure a list of preferred patient identifier systems.

reinkrul: We want to be international and BSN is very Dutchy.

dwelman: Yes good point!

reinkrul: Made it configurable, now defaults to BSN.

Note tht the line you initially pointed out can't fail, because on initialization it checks whether it can resolve the patient identifier.

PR #460, opened 2025-03-04
INT-611: Introduce extendable messaging service

reinkrul:
This introduces a `messaging` service, which lays the groundwork for reliable event and FHIR subscription notification handling. It can be extended with more topics/queues to be published on and message receivers. Additional changes:

- Mostly reimplements the existing Azure Service Bus integration for delivery of the enrolment bundle (renamed to "task filler/task accepted bundle" since Task workflows aren't just for enrolments).
- Removes the "debug" (write messages to disk) feature of the ServiceBus integration, since it was used during development, but not afterwards. Use unit tests, or the HTTP messaging endpoint instead.
- Removes the "ping" (sends a text message "ping" to a topic on startup) feature of the ServiceBus integration, since it was used during development, but will cause increased ops-load on test/acc/production environments (since it's a message that most probably can't be processed, it'll end up in the DLQ).
- Moved the delivery of enrolment bundles over HTTP (a.k.a. "demo client") to a separate HTTP message broker implementation.
- Moved configuration of message broker from CPC to ORCA root, since it'll also be used by CPS.
- Removed the Azure Service Bus setup in the local deployment, introduced an integration test instead. Reason: it starts slowly and adds additional overhead to the (already) heavy-weight local-dev deployment. We'll support an in-memory broker very soon, which is to be preferred for local-dev deployment.
- Improved documentation


orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -141,10 +141,12 @@ func (s *Service) acceptPrimaryTask(ctx context.Context, cpsClient fhirclient.Cl
 		return fmt.Errorf("failed to update primary Task status (id=%s): %w", ref, err)
 	}
 	log.Ctx(ctx).Debug().Msgf("Successfully accepted task (ref=%s)", ref)
-	err = s.notifier.NotifyTaskAccepted(ctx, cpsClient, primaryTask)
-	if err != nil {
-		log.Ctx(ctx).Warn().Msgf("Accepted Task with an error in the notification (ref=%s): %s", ref, err.Error())
-		return nil
+	if s.notifier != nil {
```
dwelman, line 144: Do we want to error out or at least log an error if notifier is nil? Considering how important this is to the task filling workflow, unless I am missing a use case where it isn't a problem to not have a notifier?

viewer_simulator/app/(DashboardLayout)/components/bgz/bgz-overview.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
-import { Bundle, BundleEntry, CarePlan, Task } from 'fhir/r4';
+import { BundleEntry, CarePlan } from 'fhir/r4';
 import CarePlanTable from './bgz-careplan-table';
 import { getBsn } from '@/utils/fhirUtils';
 import { headers } from 'next/headers'
-import { getNotificationBundles } from '@/app/api/delivery/storage';
+import { getNotificationBundles } from '@/app/api/delivery/orca-enroll-patient/storage';
```
dwelman, line 6: This might need to be refactored based on @JorisHeadease's latest change

orchestrator/messaging/azservicebus_integration_test.go
```diff
@@ -0,0 +1,123 @@
...
+	ctx := context.Background()
+	req := tc.ContainerRequest{
+		Image:        "mcr.microsoft.com/mssql/server:2022-latest",
+		ExposedPorts: []string{"1433/tcp"},
+		Networks:     []string{dockerNetwork.Name},
+		Env: map[string]string{
+			"ACCEPT_EULA":       "Y",
+			"MSSQL_SA_PASSWORD": "Z4perS3!cr3!t",
+		},
+		// WaitingFor: wait.ForLog("SQL Server is now ready for client connections"),
```
dwelman, line 60: Is this comment needed considering we have the sleep in place?

orchestrator/messaging/azservicebus_integration_test.go
```diff
@@ -0,0 +1,123 @@
...
+	// When the container signals ready, the Service Bus emulator actually isn't ready yet.
+	// See https://github.com/Azure/azure-service-bus-emulator-installer/issues/35
+	// It should be 0.5s, but we wait 3s to be on the safe side (to cope with slow CI).
+	time.Sleep(3 * time.Second)
+	t.Cleanup(func() {
+		// Shutdown in case shutdown test doesn't run
+		_ = broker.Close(context.Background())
+	})
+
+	t.Run("send message", func(t *testing.T) {
```
dwelman, line 36: Can we also get negative tests? Malformed body, non existing topic, non-responsive topic endpoint etc?

orchestrator/messaging/http_test.go
```diff
@@ -0,0 +1,55 @@
...
+	broker := HTTPBroker{
+		endpoint: testServer.URL,
+	}
+
+	message := &Message{
+		Body:        []byte(`{"key":"value"}`),
+		ContentType: "application/json",
+	}
+
+	t.Run("ok", func(t *testing.T) {
```
dwelman, line 45: Negative testing?

orchestrator/careplancontributor/handle_taskfiller.go
reinkrul: The notifier is only configured when the topic on which an enrolment bundle should be delivered, is set. So when it's not there, it's not an error but just how the environment is configured. In that case, the EHR should integrate in some different way (e.g. listening to changes in their FHIR store).

dwelman (approved):

PR #464, opened 2025-03-05
INT-627 | Implement unmanaged operations

dwelman:
Implement Create and Update (with Create via PUT supported) for Patient, ServiceRequest, Questionnaire, and QuestionnaireResponse. Auth is handled via AuditEvents


dwelman:
CRUD integ test has been fixed and updated, it is missing GET and SEARCH, I will add these tomorrow

orchestrator/careplanservice/util.go
```diff
@@ -126,9 +126,7 @@ func (s *Service) filterAuthorizedPatients(ctx context.Context, patients []fhir.
 func handleSearchResource[T any](ctx context.Context, s *Service, resourceType string, queryParams url.Values, headers *fhirclient.Headers) ([]T, *fhir.Bundle, error) {
 	form := url.Values{}
 	for k, v := range queryParams {
-		for _, value := range v {
-			form.Add(k, value)
-		}
+		form.Add(k, strings.Join(v, ","))
```
reinkrul, line 129: is this right?

orchestrator/careplanservice/integration_audit_test.go
```diff
@@ -426,8 +395,46 @@ func verifyAuditEvents(t *testing.T, fhirClient fhirclient.Client, expectedEvent
...
+			if err != nil {
+				return fmt.Errorf("failed to perform specific search for audit event: %w", err)
+			}
+
+			// Check if we found it in the specific search
+			found := false
+			for _, entry := range specificBundle.Entry {
+				var auditEvent fhir.AuditEvent
+				if err := json.Unmarshal(entry.Resource, &auditEvent); err != nil {
+					continue
```
reinkrul, line 417: you could fail the test here? Because it should always be an unmarshalable AuditEvent?

orchestrator/careplanservice/handle_getservicerequest_test.go
```diff
@@ -55,7 +78,7 @@ func TestService_handleGetServiceRequest(t *testing.T) {
 		"error: ServiceRequest exists, fetched task, incorrect principal": {
 			context: auth.WithPrincipal(context.Background(), *auth.TestPrincipal3),
 			expectedError: &coolfhir.ErrorWithCode{
-				Message:    "Participant does not have access to ServiceRequest",
+				Message:    "No creation audit event found for ServiceRequest",
```
reinkrul, line 81: I don't think we want to disclose this detail to the client, and only want to log it. That applies to all detailed messages about why authorization failed i.m.o.?

orchestrator/careplanservice/handle_getquestionnaireresponse.go
```diff
@@ -26,6 +26,15 @@ func (s *Service) handleGetQuestionnaireResponse(ctx context.Context, id string,
...
 	// If the user has access to the task, they have access to the questionnaire response
 	if len(bundle.Entry) == 0 {
+		// If the user created the questionnaire response, they have access to it
+		isCreator, err := s.isCreatorOfResource(ctx, "QuestionnaireResponse", id)
+		if err != nil {
+			return nil, err
+		}
+		if isCreator {
+			return &questionnaireResponse, nil
+		}
```
reinkrul, line 36: isn't this check in the wrong control flow block? Since `len(bundle.Entry) == 0`, indicates no access to the related tasks?

reinkrul (changes requested):

orchestrator/careplanservice/handle_createpatient.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, fmt.Errorf("invalid %T: %w", patient, coolfhir.BadRequestError(err))
+	}
+
+	// Check we're only allowing secure external literal references
+	if err := s.validateLiteralReferences(ctx, &patient); err != nil {
+		return nil, err
+	}
+
+	// Patient.id must be generated by the backing FHIR API, so it must be nil in the request
+	patient.Id = nil
```
reinkrul, line 31: Does this change existing behavior? Could it be that we didn't clear the ID for unmanaged creates earlier?

This might break St. Antonius' integration, since they rely on the patient ID they feed in the system to check if it already exists (or perform careplan-queries later on). I agreed that we won't change that for now, for the sake of removing blockades for going live.

@bramwesselo what do you think?

orchestrator/careplanservice/handle_createpatient.go
```diff
@@ -0,0 +1,83 @@
...
+	principal, err := auth.PrincipalFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get local identity for audit
+	localIdentity, err := s.getLocalIdentity()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get local identity: %w", err)
+	}
```
reinkrul, line 43: I believe we need these for most, if not all, operations. It might be an idea to put them into the FHIRHandlerRequest struct?

reinkrul:
I think there's quite some code duplication that can be fixed by introducing generic handlers for the supported FHIR operations (read, search, create, update), which can then be enables as triples; `[ resource type; operation; authpolicy ]`. Auth policy would be a function like `isCreatorOfResource`. And we might need to pass an alteration function (e.g. to set the `ServiceRequest.requester` field).

We don't need to do this now part of this PR (it's already large enough as it is, and it's smart to separate functional changes from refactoring), but maybe we can look into it somewhere soon.

orchestrator/careplanservice/util.go
dwelman: Yes, I discovered during testing that FHIR expects the query params in the form `_id=1,2,3` and not `_id=1&_id=2` as I was not getting back all the search results I expected as it previously was. This fixed my problem

orchestrator/careplanservice/integration_audit_test.go
dwelman: Fair point

orchestrator/careplanservice/handle_getservicerequest_test.go
dwelman: Yes that's a good point, definitely an oversight on my behalf. I think I will log that there is no creation event and still return the previous error

orchestrator/careplanservice/handle_getquestionnaireresponse.go
dwelman: The logic is right, the comment is wrong, I will update.

Anyone that has access to the task has access to the QR, but if you do not have access to the Task but are the creator of the QR you can still access the QR

dwelman:
@reinkrul yes I agree that a refactor for the audit events sounds reasonable, I think this is one of those cases where it was fine as I was implementing but now taking a step back it is a lot more code than I expected.

We need to implement additional audit events for errors, as well as for notifications, so this will already be a good space to do some cleanup. We will also switch our auth to OPA at some point so I didn't want to put too much work into audits as validation so let's see what the timeline of that is like

orchestrator/careplanservice/util.go
dwelman: I have added a new integration test to demonstrate this is working as expected

reinkrul:
> @reinkrul yes I agree that a refactor for the audit events sounds reasonable, I think this is one of those cases where it was fine as I was implementing but now taking a step back it is a lot more code than I expected.
> 
> We need to implement additional audit events for errors, as well as for notifications, so this will already be a good space to do some cleanup. We will also switch our auth to OPA at some point so I didn't want to put too much work into audits as validation so let's see what the timeline of that is like

Just wondering, do you think, this is a typical trait of the state of AI for coding as it currently is? Like, easy to extend existing/add new functionality, but unable to refactor existing code to avoid code duplication?

orchestrator/careplanservice/service.go
```diff
@@ -101,14 +101,16 @@ type Service struct {
...
-	Context      context.Context
+	ResourceId    string
+	ResourcePath  string
+	ResourceData  json.RawMessage
+	HttpMethod    string
+	HttpHeaders   http.Header
+	RequestUrl    *url.URL
+	FullUrl       string
+	Context       context.Context
+	Principal     *auth.Principal
```
reinkrul, line 112:
```suggestion
	// Principal contains the identity of the client invoking the FHIR operation.
	Principal     *auth.Principal
```

orchestrator/careplanservice/service.go
```diff
@@ -101,14 +101,16 @@ type Service struct {
...
+	LocalIdentity *fhir.Identifier
```
reinkrul, line 113:
```suggestion
	// LocalIdentity contains the identifier of the local care organization handling the FHIR operation invocation.
	LocalIdentity *fhir.Identifier
```

orchestrator/careplanservice/handle_updatequestionnaire.go
```diff
@@ -48,28 +45,20 @@ func (s *Service) handleUpdateQuestionnaire(ctx context.Context, request FHIRHan
 		return s.handleCreateQuestionnaire(ctx, request, tx)
 	}
 
-	// Extract the existing Questionnaire from the bundle
+	// As long as the user is authorised, they may update the questionnaire
```
reinkrul, line 48: I think only the creator is allowed to?

orchestrator/careplanservice/handle_getservicerequest.go
```diff
@@ -29,17 +30,17 @@ func (s *Service) handleGetServiceRequest(ctx context.Context, id string, header
 	if err != nil {
 		return nil, err
 	}
-	// If bundle is empty, the user does not have access to the ServiceRequest
+	// If the user does not have access to the Task, check if they are the creator of the ServiceRequest
```
reinkrul, line 33: Ah right, now I understand. Access is allowed if either:
- The principal has access to the Task (through CareTeam or being Task.request/owner), or
- The principal is the creator of the resource

orchestrator/careplanservice/handle_getquestionnaireresponse.go
```diff
@@ -19,21 +21,27 @@ func (s *Service) handleGetQuestionnaireResponse(ctx context.Context, id string,
...
 
 	// Fetch tasks where the QuestionnaireResponse is in the task Output
+	// If the user has access to the task, they have access to the questionnaire response
 	bundle, err := s.handleSearchTask(ctx, url.Values{"output-reference": []string{"QuestionnaireResponse/" + id}}, headers)
 	if err != nil {
 		return nil, err
 	}
 
-	// If the user has access to the task, they have access to the questionnaire response
+	principal, err := auth.PrincipalFromContext(ctx)
```
reinkrul, line 30: hmm, why don't we use the `FHIRHandlerRequest` struct for read operation handlers? 🤔

dwelman:
> Just wondering, do you think, this is a typical trait of the state of AI for coding as it currently is? Like, easy to extend existing/add new functionality, but unable to refactor existing code to avoid code duplication?

It's definitely possible to do refactoring but it is more deliberate and delicate, easier to write out new code. That being said, the same could be said about programming by hand :P I definitely went a bit too fast with these new tools, will be more deliberate and thoughtful going forward

orchestrator/careplanservice/handle_updatequestionnaire.go
dwelman: I was running into issues with the way the task filler is implemented, spoke to Bram about it and he said this is fine since there is no sensitive data in a questionnaire and this is just a stopgap implementation until we implement OPA anyway

orchestrator/careplanservice/handle_getquestionnaireresponse.go
dwelman: Hmm, that is a good question. I think we had implemented Get methods before we thought to use the FHIRHandlerRequest struct, and maybe it wasn't implemented for them since they don't need a transaction? Either way I think it is a good idea to standardize, I will make the change

dwelman: I'll refactor this in another PR otherwise this one will end up being even bigger than it already is

orchestrator/careplanservice/handle_createpatient.go
dwelman: Coordinated with Harmony and updated the Create methods to support create via PUT if an ID is supplied. Let me know if you want it to be more explicit for example making it a requirement to pass a flag to explicitly say you want a PUT, right now it is automatically determined via if the resource has an ID or not

dwelman: I've made it a bit smarter, it will only PUT if the FHIRRequest HTTP method is PUT

orchestrator/careplanservice/handle_createpatient.go
```diff
@@ -0,0 +1,76 @@
...
+			Type:      to.Ptr("Patient"),
+		},
+		&fhir.Reference{
+			Identifier: request.LocalIdentity,
+			Type:       to.Ptr("Organization"),
+		},
+	)
+
+	// If patient has an ID, treat as PUT operation
+	if patient.Id != nil && request.HttpMethod == "PUT" {
```
reinkrul, line 54: this looks really weird, since FHIR supports upserting (update becomes create), but vice versa, I don't think so?

orchestrator/careplanservice/handle_createpatient.go
```diff
@@ -0,0 +1,76 @@
...
+			Method: fhir.HTTPVerbPUT,
+			Url:    "Patient/" + *patient.Id,
+		}, nil, coolfhir.WithFullUrl("Patient/"+*patient.Id))
+		createAuditEvent.Entity[0].What.Reference = to.Ptr("Patient/" + *patient.Id)
+	} else {
+		tx.Create(patient, coolfhir.WithFullUrl(*patientBundleEntry.FullUrl))
+	}
+
+	patientEntryIdx := len(tx.Entry) - 1
+	tx.Create(createAuditEvent)
```
reinkrul, line 65: didn't you add the "WithAuditEvent()" function?

orchestrator/careplanservice/handle_createquestionnaire.go
```diff
@@ -0,0 +1,69 @@
...
+			Type:      to.Ptr("Questionnaire"),
+		},
+		&fhir.Reference{
+			Identifier: request.LocalIdentity,
+			Type:       to.Ptr("Organization"),
+		},
+	)
+
+	// If questionnaire has an ID, treat as PUT operation
+	if questionnaire.Id != nil && request.HttpMethod == "PUT" {
```
reinkrul, line 47: same question as for the patient resource

orchestrator/careplanservice/handle_updatepatient.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Check we're only allowing secure external literal references
+	if err := s.validateLiteralReferences(ctx, &patient); err != nil {
+		return nil, err
+	}
+
+	// Search for the existing Patient
+	var searchBundle fhir.Bundle
+	patientId := ""
+	if patient.Id != nil {
```
reinkrul, line 32: wasn't this more elaborate earlier, with it also handling upsert with logical identifiers?
E.g.

```
PUT Patient?bsn=1234
```

orchestrator/careplanservice/handle_updatequestionnaireresponse.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/SanteonNL/orca/orchestrator/lib/to"
+	"github.com/rs/zerolog/log"
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+func (s *Service) handleUpdateQuestionnaireResponse(ctx context.Context, request FHIRHandlerRequest, tx *coolfhir.BundleBuilder) (FHIRHandlerResult, error) {
+	log.Ctx(ctx).Info().Msgf("Updating QuestionnaireResponse: %s", request.RequestUrl)
+
+	// Log the raw request data for debugging
+	log.Ctx(ctx).Debug().Msgf("resourceData: %s", string(request.ResourceData))
```
reinkrul, line 21: probably want to remove this or implement it in generic middleware

orchestrator/careplanservice/service.go
```diff
@@ -419,6 +437,48 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
 		DoAndWrite(httpResponse, resource, http.StatusOK)
 }
 
+func (s *Service) handleCreate(resourcePath string) func(context.Context, FHIRHandlerRequest, *coolfhir.BundleBuilder) (FHIRHandlerResult, error) {
+	resourceType := getResourceType(resourcePath)
+
+	switch resourceType {
+	case "Task":
```
reinkrul, line 444: aren't we missing Condition?

reinkrul (changes requested):

orchestrator/careplanservice/handle_updatepatient.go
dwelman: It was previously handled as an unmanaged operation, maybe you are thinking of updateTask? 

If we need it to be more elaborate that's fine and it can be implemented but I'd need to know the full requirements and it's out of scope for this specific task

orchestrator/careplanservice/service.go
dwelman: Condition was not a requirement for INT-627, but if it is needed I can coordinate with Bram on access requirements

orchestrator/careplanservice/handle_createpatient.go
dwelman: Or maybe a flag in the FHIR request? `createWithSuppliedId` or something? Then we can set that when calling Create from Update, but it will not be set when Create is called directly from a POST

reinkrul: > Or do we just allow Create with an ID as long as it doesn't already exist?
This is the easiest solution, I think. And I don't think it'll break anything currently.

reinkrul (approved):

PR #470, opened 2025-03-07
INT-642 > INT-655: Create Observations from the MSC viewer

JorisHeadease:
This PR adds the following:

1. The order Tasks that come from the hospitals now have a "pencil" icon in the "Onboarding Tasks" table.
Screenshot 2025-03-07 at 15 13 57

2. When clicking this pencil, a popup is shown that allows for creation of a FHIR `Observation`
Screenshot 2025-03-07 at 15 14 42

3. Rewrote the FHIR endpoint, that used to proxy to become a light-weight in-memory FHIR server. This to keep the configuration simple, and to not re-introduce a second HAPI server like we used to have.
4. The Observation is stored in the new light-weight FHIR server. The Observation can be queried here as well.
Screenshot 2025-03-07 at 15 17 23

5. Also added a "fetch resources when there are no notifications in-memory" logic when the viewer is started from the dev server (`process.env.NODE_ENV === 'development'`). As the dev server does many hot reloads, the in-memory data is often cleared. This would require devs to keep moving back to the EHR simulator and create new enrolments.
6. Fixed some small dark-theme issues
7. Fixed deprecated eslint config that doesn't work in eslint 9 anymore

The `Observation` will serve as a basis for INT-642. The CPC node of the hospital should be allowed to query this MSC Observation through the SCP standard.

### DIFF
--- a/viewer_simulator/app/(DashboardLayout)/components/bgz/bgz-overview.tsx
+++ b/viewer_simulator/app/(DashboardLayout)/components/bgz/bgz-overview.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Bundle, BundleEntry, CarePlan, Task } from 'fhir/r4';
+import { BundleEntry, CarePlan } from 'fhir/r4';
 import CarePlanTable from './bgz-careplan-table';
 import { getBsn } from '@/utils/fhirUtils';
 import { headers } from 'next/headers'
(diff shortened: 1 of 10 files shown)


dwelman (approved):
Can't comment on the FE code specifically but it looks solid, this seems like a nice solution for now to show some more features and benefit while working on a more stable and solid solution in the background

viewer_simulator/app/(DashboardLayout)/components/bgz/bgz-overview.tsx
```diff
@@ -33,21 +33,25 @@ export default async function BgzOverview() {
...
+                    carePlan.contained?.find((resource: any) => resource.resourceType === "CareTeam") ??
+                    (carePlan.careTeam?.[0]?.reference ? resourceMap?.get(carePlan.careTeam[0].reference) : undefined);
 
                 if (!careTeam) {
                     console.warn(`No CareTeam found for CarePlan/${carePlan.id}`);
                 }
 
                 return {
-                    id: careTeam.id,
+                    id: careTeam?.id || "Uknown",
```
reinkrul, line 47: Unknown

PR #475, opened 2025-03-12
Allow disable notification bundle for viewer in production

dmeijboom:
Changes `getNotificationBundles` in the viewer to fetch resources instead of listening to notification bundles (only) based on the `DISABLE_NOTIFICATION_BUNDLE` environment variable instead of relying on `NODE_ENV`.

This makes the feature explicit and allows production deployments to use it.


reinkrul (changes requested):
What do we consider "production"? Since it's test software, i'd never expect this to be used in production environments.

dmeijboom:
> What do we consider "production"? Since it's test software, i'd never expect this to be used in production environments.

I was referring to environments where `NODE_ENV = production` (which is every standalone deployment in NextJS since they force the value of `NODE_ENV` to be hardcoded to `production`). It could be useful in test deployments.

Nevertheless, after talking with the rest it isn't a good idea anyway.

PR #476, opened 2025-03-12
INT-611: Make enrolment bundle collection/sending retryable

reinkrul:
This introduces the `orca.taskengine.task~accepted` queue, which allows enrolment bundle collection and sending (to the EHR) to be retried.


orchestrator/careplancontributor/ehr/notifier_test.go
```diff
@@ -80,43 +84,39 @@ func TestNotifyTaskAccepted(t *testing.T) {
...
+		// expectedSendMessageError is the error expected when sending a message to the broker.
+		expectedSendMessageError error
+		// expectedProcessMessageError is the error expected when processing a message from the broker (the message handler creating the bundle).
+		expectedProcessMessageError error
 	}{
 		{
 			name: "successful notification",
 			task: primaryTask,
-			setupMocks: func(client *test.StubFHIRClient, mockMessageBroker *messaging.MockBroker) {
+			setup: func(client *test.StubFHIRClient, messageBroker *messaging.MemoryBroker) {
```
dwelman, line 98: Nice!

orchestrator/messaging/azservicebus.go
```diff
@@ -0,0 +1,167 @@
...
+			messages, err := receiver.ReceiveMessages(c.ctx, 1, &azservicebus.ReceiveMessagesOptions{})
+			if err != nil {
+				const backoffTime = time.Minute
+				log.Ctx(c.ctx).Err(err).Msgf("AzureServiceBus: receive message failed, backing off for %s (queue: %s)", backoffTime, queue)
+				// Sleep for a minute before retrying, to avoid spamming the logs.
+				// But, the server might be instructed to shut down in the meantime, and we don't want the sleep to block the shutdown.
+				// So, use a select to also listen for shutdown.
+				select {
+				case <-c.ctx.Done():
+					return
```
dwelman, line 121: What happens to the message if we receive a shutdown while waiting for a retry?

reinkrul: When the application isn't actively handling the message (first delivery, or retry, doesn't matter), it'll stay on the queue. When the application is started again it'll receive that message again and will attempt to process it.

If the application is actively handling the message, the `Close()` call will block (using the `receivers` WaitGroup) until message handling is done (either completed successfully or "abandoned" a.k.a. put back on the queue). When the receiver `for`-loop exits (because the context was cancelled), it decreases the waitgroup and `Close()` will unblock.

dwelman: Cool, sounds good

orchestrator/messaging/service.go
```diff
@@ -12,10 +12,15 @@ func New(config Config, topics []string) (Broker, error) {
...
 	if config.AzureServiceBus.ConnectionString != "" || config.AzureServiceBus.Hostname != "" {
-		broker, err = newAzureServiceBusBroker(config.AzureServiceBus, topics)
+		broker, err = newAzureServiceBusBroker(config.AzureServiceBus, topics, config.TopicPrefix)
 		if err != nil {
 			return nil, fmt.Errorf("azure service bus: %w", err)
 		}
+	} else {
+		// If no configuration is provided, default to an in-memory broker
+		log.Warn().Msg("No messaging configuration provided, defaulting to in-memory broker. " +
+			"This is unsuitable for production use, since failed messages can't be retried.")
```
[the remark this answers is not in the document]
reinkrul: Hmm, we only use it for EHR integration right now, which only applies to ZBJ. So it doesn't apply to the hospital deployments, but it will, very soon. I'll add in the code (already wrote it), but comment it out for now.

PR #477, opened 2025-03-12
INT-611: Allow filtering of topics enabled for sending to HTTP endpoint

reinkrul:
We'll be having more topics soon, so we need to be able to filter.

### DIFF
--- a/orchestrator/messaging/http.go
+++ b/orchestrator/messaging/http.go
@@ -15,8 +15,9 @@ import (
 var _ Broker = &HTTPBroker{}
 
 type HTTPBrokerConfig struct {
-	Endpoint string   `koanf:"endpoint"`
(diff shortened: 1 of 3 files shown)


orchestrator/messaging/http.go
```diff
@@ -27,6 +29,7 @@ func NewHTTPBroker(config HTTPBrokerConfig, underlyingBroker Broker) Broker {
 type HTTPBroker struct {
 	underlyingBroker Broker
 	endpoint         string
+	topics           []string
```
dwelman, line 32: Maybe we could make it more explicit that this is a topic filter? Or at least a comment saying so?

dwelman (approved):

dwelman (approved):

PR #493, opened 2025-03-17
INT-672 | Add audit event option to Bundle/transaction builder

dwelman:
This is a small part of a larger task to do audit event handling as part of a transaction

### DIFF
--- a/orchestrator/lib/coolfhir/bundle.go
+++ b/orchestrator/lib/coolfhir/bundle.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	fhirclient "github.com/SanteonNL/go-fhir-client"
 	"github.com/SanteonNL/orca/orchestrator/lib/to"
@@ -27,6 +28,13 @@ func SearchSet() *BundleBuilder {
 	return &BundleBuilder{
 		Type: fhir.BundleTypeSearchset,
 	}
+}
+
+func (t *BundleBuilder) Get(resource interface{}, path string, opts ...BundleEntryOption) *BundleBuilder {
+	return t.Append(resource, &fhir.BundleEntryRequest{
+		Method: fhir.HTTPVerbGET,
+		Url:    path,
+	}, nil, opts...)
 }
 
 func (t *BundleBuilder) Create(resource interface{}, opts ...BundleEntryOption) *BundleBuilder {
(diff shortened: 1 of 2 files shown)


orchestrator/lib/coolfhir/bundle.go
```diff
@@ -54,29 +62,204 @@ func (t *BundleBuilder) Append(resource interface{}, request *fhir.BundleEntryRe
...
+					Type:       to.Ptr("Device"),
+				},
+			},
+			Entity: []fhir.AuditEventEntity{
+				{
+					What: &resourceRef,
+				},
+			},
+			Meta: &fhir.Meta{
+				VersionId: to.Ptr("1"),
```
reinkrul, line 231: I think this is managed/set by the FHIR server

orchestrator/lib/coolfhir/bundle.go
```diff
@@ -29,6 +30,13 @@ func SearchSet() *BundleBuilder {
 	}
 }
 
+func (t *BundleBuilder) Get(resource interface{}, path string, opts ...BundleEntryOption) *BundleBuilder {
```
reinkrul, line 33: do we require this on GET operations as AuditEvents? It's now written into the CPS. Is that the right store for these audits?

(did we align these requirements properly)

orchestrator/lib/coolfhir/bundle.go
dwelman: Ah fair point, I will remove this

orchestrator/lib/coolfhir/bundle.go
dwelman: I did align with Bram that we want audits on Get and Search, this is future groundwork because currently our GET methods are not handled as transactions but we want to move to that so the audit events are created properly.

This will be implemented in a future PR.

Am I understanding your question correctly?

orchestrator/lib/coolfhir/bundle.go
```diff
@@ -54,29 +62,204 @@ func (t *BundleBuilder) Append(resource interface{}, request *fhir.BundleEntryRe
...
+			interactionCode = "create"
+			interactionDisplay = "Create"
+		case fhir.AuditEventActionR:
+			interactionCode = "read"
+			interactionDisplay = "Read"
+		case fhir.AuditEventActionU:
+			interactionCode = "update"
+			interactionDisplay = "Update"
+		default:
+			// Considering we decide when to call this function, this will likely never happen, but it's good to have it here
```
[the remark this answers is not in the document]
dwelman: To have this throw an error I'd need to refactor all the bundle methods to return an error and all the places in which we are using bundles, which will end up touching quite a few points in the codebase. That's fine with me, if that's fine with you I can do so.

Otherwise, `ExecuteTransaction` and `NormalizeTransactionBundleResponseEntry` already have error handling built in, maybe I can check for a `FailedBundleEntry` before executing the transaction and returning an error there?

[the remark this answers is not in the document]
dwelman: I have added error handling for this to ExecuteTransaction, I think that is the cleanest way to handle it

reinkrul (approved):

PR #495, opened 2025-03-17
Basic FHIR rego access policy

dmeijboom:
## Todo

[ ] Fix unittests
...More?


reinkrul:
What were the considerations of embedding OPA v.s. invoking a standalone OPA instance?

reinkrul:
I'll be missing context for when/if I need to review this. If that's important, maybe you can document the use case(s) and how they're implemented in ORCA/Orchestrator in a/the README file?

reinkrul:
General remarks:
- This only authorizes data access on the CarePlanService. It doesn't expose data (e.g. the BgZ) from the EHR, which is exposed through the CarePlanContributor's FHIR endpoint. Wasn't that in scope?

orchestrator/careplanservice/policy/agent.go
```diff
@@ -0,0 +1,199 @@
...
+	)
+	query, err := r.PrepareForEval(ctx)
+	if err != nil {
+		return Agent{}, err
+	}
+
+	return Agent{
+		query:  query,
+		client: client,
+		logger: zerolog.New(os.Stdout),
```
reinkrul, line 86: can you derive the logger from `ctx`? Then contextual log fields are added as well.

orchestrator/careplanservice/policy/agent.go
```diff
@@ -0,0 +1,199 @@
...
+	result, err := m.query.Eval(ctx, rego.EvalInput(context))
+	if err != nil {
+		return fmt.Errorf("failed to evaluate policy: %w", err)
+	}
+
+	if len(result) == 0 {
+		return fmt.Errorf("policy evaluation failed")
+	}
+
+	allow, ok := result[0].Bindings["allow"]
```
reinkrul, line 100: what if there's multiple results? If we only expect/support 1, I'd fail if there are multiple (to avoid giving access when we shouldn't)

orchestrator/careplanservice/policy/policy.rego
```diff
@@ -0,0 +1,87 @@
...
+is_careteam_participant if {
+	some care_plan in input.careplans
+	care_teams := extract_care_teams(care_plan)
+	some care_team in care_teams
+	some participant in care_team.participant
+	identifier_matches(participant.member.identifier, input.principal)
+}
+
+is_get_post if {
+	input.method in ["GET", "POST"]
```
reinkrul, line 53: we also support PUT. And, you could map it to FHIR operations (read, search, update, create) if you want.

orchestrator/careplanservice/policy/policy.rego
```diff
@@ -0,0 +1,87 @@
...
+}
+
+default allow := false
+
+allow if {
```
reinkrul, line 58: do all of these `allow` if need to be `true` to have `allow` evaluate to `true`, or one of them?

orchestrator/careplanservice/policy/policy.rego
```diff
@@ -0,0 +1,87 @@
...
+	input.method == "GET"
+	input.resource.resourceType == "Questionnaire"
+}
+
+allow if {
+	input.method == "GET"
+	input.resource.resourceType == "QuestionnaireResponse"
```
reinkrul, line 65: I believe `QuestionnaireResponse` has the same authz rules (for `read`) as the `Patient` resource, as it contains medical data.

orchestrator/careplanservice/policy/policy.rego
```diff
@@ -0,0 +1,87 @@
...
+}
+
+allow if {
+	is_get_post
+	is_careteam_participant
+	not contains_tags
```
reinkrul, line 71: ?

orchestrator/careplanservice/policy/agent.go
```diff
@@ -0,0 +1,199 @@
...
+
+	rolesHeader := r.Header.Get("Orca-Auth-Roles")
+	if rolesHeader != "" {
+		roles = strings.Split(rolesHeader, ",")
+	}
+
+	return &Preflight{
+		ResourceId:   id,
+		ResourceType: resourceType,
+		Query:        r.URL.Query(),
```
reinkrul, line 142: We only support POST-based FHIR searches to avoid leaking PII in logging, then the search parameters need to be read from the HTTP request body.

But, you can do a conditional update/create in FHIR in which you _do_ specify search parameters in the HTTP request URL, but we might've removed support for that recently (@dwelman ?); e.g.:

```http
PUT /Task?ehr.identifier=1234
{"resourceType": "Task"}
```

(updates the Task identified by this logical identifier)

reinkrul (changes requested):
initial review

orchestrator/careplanservice/policy/policy.rego
dmeijboom: There is no logical OR in Rego so instead I've duplicated the 'behandelrelatie' rule into two separate rules:
1. Without tags
2. With tags, assert tags

orchestrator/careplanservice/policy/policy.rego
dmeijboom: One of them, it's basically a logical OR

orchestrator/careplanservice/service.go
```diff
@@ -89,13 +97,21 @@ func New(config Config, profile profile.Provider, orcaPublicURL *url.URL, messag
...
 	}
 	return &s, nil
 }
 
+//go:generate mockgen -destination=./policy_agent_mock.go -package=careplanservice -source=service.go PolicyAgent
+
+type PolicyAgent interface {
+	Allow(ctx context.Context, context *policy.Context) error
+	Preflight(resourceType, id string, r *http.Request) (*policy.Preflight, error)
+	PrepareContext(ctx context.Context, cache policy.SearchCache, preflight *policy.Preflight, resource any) (*policy.Context, error)
```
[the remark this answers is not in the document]
dmeijboom: Basically because of Search. In that case we are doing one `Preflight` call with multiple `PrepareContext` calls

orchestrator/careplanservice/policy/policy.rego
bramwesselo: It is; a so called POST-search can be used to retrieve data (POST {{FHIR_URL}}/Patient/_search) parameter go in the body (form-encoded)

orchestrator/careplanservice/service.go
```diff
@@ -89,13 +97,21 @@ func New(config Config, profile profile.Provider, orcaPublicURL *url.URL, messag
...
+	err = s.ensureCustomSearchParametersExists(ctx)
 	if err != nil {
 		return nil, err
 	}
 	return &s, nil
 }
 
+//go:generate mockgen -destination=./policy_agent_mock.go -package=careplanservice -source=service.go PolicyAgent
+
+type PolicyAgent interface {
```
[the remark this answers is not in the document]
dmeijboom: This follows best practises in Go where it's common to define the interface at the consumer. But, I have to agree that it doesn't fit in with other packages in ORCA (and many other Go repos). So, I'll move it to the `policy` package.

orchestrator/careplanservice/policy/policy.rego
bramwesselo: Agree. Questionnaire does not relate to a Patient, QuestionnaireResponse does

joostholslag:
@reinkrul > I think for authorization, we should maintain 1 implementation for the "core" business rules of SCP (since they should be built-in

Why should they be built in? Externalising them (to OPA/REGO) has major benefits. Or what do you mean with 'built in'?

bramwesselo:
> * Do you think we should go through with OPA, with the goal of ultimately replacing all authorization logic by it?
>   
>   * I was told it was only supplemental by @bramwesselo, not a replacement for our SCP authorization logic. I'm not against replacing it, if we all agree that we can maintain just as well as our existing authz logic

Last week, I had a discussion with @dmeijboom on what we think of OPA/Rego now we have some experience.  Long story short: just using Golang might be more efficient for Orca. I don't think using OPA/Rego is ever going to be mandated by anybody. I do think OPA/Rego is good to specify (legal) authorization logic and it should be used as reference implementation. The OPA/Rego-unittests (inputs and result) should be copied/adopted in Orca.

bramwesselo:
> @reinkrul
> 
> > I think for authorization, we should maintain 1 implementation for the "core" business rules of SCP (since they should be built-in
> 
> Why should they be built in? Externalising them (to OPA/REGO) has major benefits. Or what do you mean with 'built in'?

@joostholslag: Our current view is that the benefits of 'built-in' authorization logic outweigh the benefits of externalizing them. As I said in my last comment; I do think OPA/Rego is good to specify (legal) authorization logic and it should be used as a reference implementation. The OPA/Rego-unittests (inputs and result) should be copied/adopted in Orca (or any other healthcare system in NL).

As to why this is our current view:
- We think Rego should reflect/implement legal regulations. This legislation does not change frequently
- Authorization logic is pretty concise & simple; Basically "allow if Treatment-relationship  or Consent" and some added filters or pseudomisation based on the role of the user/system. It is not a lot of work to copy this logic to your own programming language.
- The application that uses the OPA-instance/Rego-script should implement the desired Rego-inputs or expected Rego-outputs; that means having to deal with an extra language (Rego), host an extra container (OPA) and managing changes and dependencies. 
- Autorization logic may be applied to the incoming request (e.g. allow/disallow or search-narrowing) or on the response (filter response data based on the role of the user (a.g.v. rolzib-matrix)). You could support both flavors in the Rego-script, but you might as well leave that (and/or other application specific/performance optimizations) to the actual application/EHR/ECD/HIS-developers.

PR #507, opened 2025-03-21
INT-668: Removed the unused EHR client from the frontend.

JorisHeadease:


dwelman (approved):

reinkrul (approved):
Less code, always good!

PR #535, opened 2025-03-27
INT-699: Upgrade the `frontend` dependencies

JorisHeadease:
This PR attempt to go over all dependencies, and use the latest versions where possible. Initially minor upgrades, then major upgrades (where possible)


JorisHeadease:
Performed minor and major updates. With the exception of `@mui/material` to v7 as `@aehrc/smart-forms-renderer` is still using v6. The UI will break when upgrading to v7.

PR #536, opened 2025-03-27
INT-686 | Refactor Read methods with Audit creation

dwelman:
Update CRUD methods to support multiple results being returned by the transaction handler, to support Search transactions.

Implement transactions for Get and Search.

----

TODO: 
Proxy tests are failing due to tests being written on old assumptions about GET, busy fixing these
Remove unmanaged operations


orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -2,23 +2,28 @@ package careplanservice
...
+	"github.com/rs/zerolog/log"
 	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
 )
 
 // handleGetCarePlan fetches the requested CarePlan and validates if the requester has access to the resource (is a participant of one of the CareTeams of the care plan)
 // if the requester is valid, return the CarePlan, else return an error
-// Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
-func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhirclient.Headers) (*fhir.CarePlan, error) {
+func (s *Service) handleGetCarePlan(ctx context.Context, request FHIRHandlerRequest, tx *coolfhir.BundleBuilder) (FHIRHandlerResult, error) {
+	log.Ctx(ctx).Info().Msgf("Getting CarePlan with ID: %s", request.ResourceId)
```
reinkrul, line 22: I'd name these more according to FHIR, which is a `read` operation in this case

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -28,29 +33,112 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+		},
+		Observer: *request.LocalIdentity,
+	}))
 
-	return &carePlan, nil
+	carePlanEntryIdx := 0
+
+	return func(txResult *fhir.Bundle) ([]*fhir.BundleEntry, []any, error) {
+		var retCarePlan fhir.CarePlan
+		result, err := coolfhir.NormalizeTransactionBundleResponseEntry(ctx, s.fhirClient, s.fhirURL, &tx.Entry[carePlanEntryIdx], &txResult.Entry[carePlanEntryIdx], &retCarePlan)
```
reinkrul, line 54: we don't use `retCarePlan`, can we pass in `nil`?

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -28,29 +33,112 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+
+	return func(txResult *fhir.Bundle) ([]*fhir.BundleEntry, []any, error) {
+		results := []*fhir.BundleEntry{}
+
+		for _, idx := range carePlanEntryIndexes {
+			var retCarePlan fhir.CarePlan
+
+			result, err := coolfhir.NormalizeTransactionBundleResponseEntry(ctx, s.fhirClient, s.fhirURL, &tx.Entry[idx], &txResult.Entry[idx], &retCarePlan)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to process CarePlan read result: %w", err)
```
reinkrul, line 124:
```suggestion
				return nil, nil, fmt.Errorf("failed to process CarePlan search result: %w", err)
```

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -28,29 +33,112 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+			Observer:         *request.LocalIdentity,
+			AdditionalEntity: []fhir.AuditEventEntity{queryEntity},
+		}))
+	}
+
+	return func(txResult *fhir.Bundle) ([]*fhir.BundleEntry, []any, error) {
+		results := []*fhir.BundleEntry{}
+
+		for _, idx := range carePlanEntryIndexes {
+			var retCarePlan fhir.CarePlan
```
reinkrul, line 120: same here

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -28,29 +33,112 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+		for param, values := range request.QueryParams {
+			queryEntity.Detail = append(queryEntity.Detail, fhir.AuditEventEntityDetail{
+				Type:        param, // parameter name as string
+				ValueString: to.Ptr(strings.Join(values, ",")),
+			})
+		}
+
+		carePlanEntryIndexes = append(carePlanEntryIndexes, len(tx.Entry))
+
+		tx.Get(entry.Resource, "CarePlan/"+*currentCarePlan.Id, coolfhir.WithAuditEvent(ctx, tx, coolfhir.AuditEventInfo{
```
reinkrul, line 105: why read the CarePlan again if we already have it? Can't we just write the audit events?

orchestrator/careplanservice/handle_getcareplan.go
```diff
@@ -28,29 +33,112 @@ func (s *Service) handleGetCarePlan(ctx context.Context, id string, headers *fhi
...
+		// We do not want to notify subscribers for a get
+		return results, []any{}, nil
+	}, nil
+}
+
+// searchCarePlan performs the core functionality of searching for care plans and filtering by authorization
+// This can be used by other resources to search for care plans and filter by authorization
+func (s *Service) searchCarePlan(ctx context.Context, queryParams url.Values, headers *fhirclient.Headers, principal auth.Principal) (*fhir.Bundle, error) {
+	// Verify requester is authenticated
+	if principal.Organization.Identifier == nil || len(principal.Organization.Identifier) == 0 {
```
reinkrul, line 138: I don't think this is possible, and we only do it here? (so, why _just_ here?)

orchestrator/careplanservice/handle_getpatient.go
```diff
@@ -37,18 +38,102 @@ func (s *Service) handleGetPatient(ctx context.Context, id string, headers *fhir
 		}
 	}
 
-	return &patient, nil
+	tx.Get(patient, "Patient/"+request.ResourceId, coolfhir.WithAuditEvent(ctx, tx, coolfhir.AuditEventInfo{
```
reinkrul, line 41: why read again?

orchestrator/careplanservice/service.go
```diff
@@ -256,8 +257,10 @@ func (s *Service) commitTransaction(request *http.Request, tx *coolfhir.BundleBu
 		if err != nil {
 			return nil, fmt.Errorf("bundle execution succeeded, but couldn't resolve bundle.entry[%d] results: %w", entryIdx, err)
 		}
-		if currResult != nil {
-			resultBundle.Entry = append(resultBundle.Entry, *currResult)
+		if len(currResult) > 0 {
```
reinkrul, line 260: no need for this if-statement, since you range over it

orchestrator/careplanservice/service.go
```diff
@@ -293,12 +296,18 @@ func (s *Service) handleUnmanagedOperation(request FHIRHandlerRequest, tx *coolf
...
 	}
 
+	idx := len(tx.Entry)
 	requestBundleEntry := request.bundleEntry()
-	tx.AppendEntry(requestBundleEntry)
-	idx := len(tx.Entry) - 1
-	return func(txResult *fhir.Bundle) (*fhir.BundleEntry, []any, error) {
+	// TODO: determine the appropriate audit event action based on the HTTP method
+	tx.AppendEntry(requestBundleEntry, coolfhir.WithAuditEvent(request.Context, tx, coolfhir.AuditEventInfo{
+		Action: fhir.AuditEventActionR,
```
reinkrul, line 303: this is missing acting agent and observer?

orchestrator/careplanservice/service.go
```diff
@@ -377,69 +387,75 @@ func (s *Service) handleModification(httpRequest *http.Request, httpResponse htt
...
+	var statusCode int
+	fhirResponse := txResult.Entry[0].Response
+	statusParts := strings.Split(fhirResponse.Status, " ")
+	if statusCode, err = strconv.Atoi(statusParts[0]); err != nil {
+		log.Ctx(httpRequest.Context()).Warn().Msgf("Failed to parse status code from transaction result (responding with 200 OK): %s", fhirResponse.Status)
+		statusCode = http.StatusOK
+	}
+	var headers = map[string][]string{}
+	if fhirResponse.Location != nil {
+		headers["Location"] = []string{*fhirResponse.Location}
```
reinkrul, line 444: I feel like I already wrote this code somewhere else

orchestrator/careplanservice/service.go
```diff
@@ -548,15 +589,18 @@ func (s *Service) handleSearch(httpRequest *http.Request, httpResponse http.Resp
...
 	case "CarePlan":
-		bundle, err = s.handleSearchCarePlan(httpRequest.Context(), queryParams, headers)
+		handler = s.handleSearchCarePlan
 	case "Task":
-		bundle, err = s.handleSearchTask(httpRequest.Context(), queryParams, headers)
-	case "Patient":
-		bundle, err = s.handleSearchPatient(httpRequest.Context(), queryParams, headers)
+		handler = s.handleSearchTask
 	default:
+		// Handle unmanaged resources
```
reinkrul, line 603: this handling differs from `handleRead`?

orchestrator/careplanservice/service.go
```diff
@@ -568,84 +612,84 @@ func (s *Service) handleSearch(httpRequest *http.Request, httpResponse http.Resp
...
-			// Add the query entity to the audit event
-			auditEvent.Entity = append(auditEvent.Entity, queryEntity)
-
-			err = s.fhirClient.Create(auditEvent, &auditEvent)
-			if err != nil {
-				coolfhir.WriteOperationOutcomeFromError(httpRequest.Context(), err, operationName, httpResponse)
-				return
-			}
+		// Add common headers if present
+		if fhirResponse.Location != nil {
```
reinkrul, line 679: duplicate code?

reinkrul (changes requested):
I think we need to think about abstracting FHIR operation handlers (create, read, search, update). Earlier we said we don't want to do this right now, because OPA policies will replace much of it. But it'll probably not replace it 100% because of its complexity and the implementations might exist side-by-side for at least some time.

That said, probably not part of this PR. I think we're making the operation handlers more and more alike to each other (that's good, makes abstraction easier), but we're still adding functionality to all of them (digging a deeper hole, that's bad).

orchestrator/careplanservice/handle_getcareplan.go
dwelman: Sounds good, will do

orchestrator/careplanservice/handle_getcareplan.go
dwelman: As discussed offline, I will change how these are returned so this will be fixed

orchestrator/careplanservice/service.go
dwelman: This will be removed since we no longer support unmanaged operations

orchestrator/careplanservice/service.go
dwelman: Right, good point

orchestrator/careplanservice/handle_getcareplan.go
dwelman: No longer relevant

reinkrul (approved):

jdksloan (approved):

PR #538, opened 2025-03-28
INT-700: Remove scrollToTop logic

JorisHeadease:
Was used before when there were multiple Questionnaires. Now simply goes to the Task status page, so it'll be at the top after a new Task is received. 

If needed, this logic can be performed on Task updates, which is cleaner anyway.


reinkrul:
Can we invite Dennis to start reviewing these PRs as well?

JorisHeadease:
Sure! Do you know his GitHub user id @reinkrul? Don't see him in the suggested reviewers

PR #568, opened 2025-04-17
FIX | Implement missing search APIs

dwelman:
The BGZ viewer does queries on all our data types using Search, previously this was handled by the proxy but this change implements Search for each type explicitly

### DIFF
--- a/orchestrator/careplanservice/handle_getcondition.go
+++ b/orchestrator/careplanservice/handle_getcondition.go
@@ -5,9 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 
 	fhirclient "github.com/SanteonNL/go-fhir-client"
 	"github.com/SanteonNL/orca/orchestrator/lib/audit"
+	"github.com/SanteonNL/orca/orchestrator/lib/auth"
 	"github.com/SanteonNL/orca/orchestrator/lib/coolfhir"
 	"github.com/SanteonNL/orca/orchestrator/lib/to"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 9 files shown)


orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -74,3 +77,116 @@ func (s *Service) handleReadCondition(ctx context.Context, request FHIRHandlerRe
 		return []*fhir.BundleEntry{&bundleEntry}, []any{}, nil
 	}, nil
 }
+
+// handleSearchCondition performs a search for Condition based on the user request parameters
+// and filters the results based on user authorization
+// Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
```
reinkrul, line 83: incorrect, useless comment?
```suggestion
// Pass in a pointer to a fhirclient.Headers object to get the headers from the fhir client request
```

orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -74,3 +77,116 @@ func (s *Service) handleReadCondition(ctx context.Context, request FHIRHandlerRe
...
+		for param, values := range request.QueryParams {
+			queryEntity.Detail = append(queryEntity.Detail, fhir.AuditEventEntityDetail{
+				Type:        param, // parameter name as string
+				ValueString: to.Ptr(strings.Join(values, ",")),
+			})
+		}
+
+		bundleEntry := fhir.BundleEntry{
+			Resource: entry.Resource,
+			Response: &fhir.BundleEntryResponse{
```
reinkrul, line 123: I doubt whether you need a response for every search result, since they're not independent operations from a transaction-type bundle (they're not individual results from individual operations, but a resultset).

orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -74,3 +77,116 @@ func (s *Service) handleReadCondition(ctx context.Context, request FHIRHandlerRe
...
+		// If there are no conditions in the bundle there is no point in doing validation, return empty bundle to user
+		return &fhir.Bundle{Entry: []fhir.BundleEntry{}}, nil
+	}
+
+	// For each Condition, verify the user has access to the patient referenced in the subject
+	var allowedConditionRefs []string
+	for _, cond := range conditions {
+		// Verify if the Condition has a valid Patient subject and the user has access to this patient
+		if cond.Subject.Identifier != nil && cond.Subject.Identifier.System != nil && cond.Subject.Identifier.Value != nil {
+			patientBundle, err := s.searchPatient(ctx, map[string][]string{"identifier": {fmt.Sprintf("%s|%s", *cond.Subject.Identifier.System, *cond.Subject.Identifier.Value)}}, headers, principal)
```
reinkrul, line 164: This leads to `handleSearchCondition` being an `N+1` (actually, N+2) query: for every matching Condition, it searches for the patient, then for the CarePlan. This could be very slow, especially if there are a lot of matching conditions. I didn't check, but I suspect the same pattern is present in other search handlers?

So maybe we need to introduce some kind of caching. But then, if it's in other search handlers as well, we might want to address at some lower level (e.g. caching built into FHIRClient). But then, it's probably out of scope for this PR.

orchestrator/careplanservice/handle_getcondition.go
```diff
@@ -74,3 +77,116 @@ func (s *Service) handleReadCondition(ctx context.Context, request FHIRHandlerRe
...
+	if len(conditions) == 0 {
+		// If there are no conditions in the bundle there is no point in doing validation, return empty bundle to user
+		return &fhir.Bundle{Entry: []fhir.BundleEntry{}}, nil
+	}
+
+	// For each Condition, verify the user has access to the patient referenced in the subject
+	var allowedConditionRefs []string
+	for _, cond := range conditions {
+		// Verify if the Condition has a valid Patient subject and the user has access to this patient
+		if cond.Subject.Identifier != nil && cond.Subject.Identifier.System != nil && cond.Subject.Identifier.Value != nil {
```
reinkrul, line 163: we might have a helper for this boolean check in coolfhir

orchestrator/careplanservice/handle_getquestionnaire.go
```diff
@@ -48,3 +50,79 @@ func (s *Service) handleReadQuestionnaire(ctx context.Context, request FHIRHandl
...
+		bundleEntry := fhir.BundleEntry{
+			Resource: entry.Resource,
+			Response: &fhir.BundleEntryResponse{
+				Status: "200 OK",
+			},
+		}
+		results = append(results, &bundleEntry)
+
+		// Add audit event to the transaction
+		auditEvent := audit.Event(*request.LocalIdentity, fhir.AuditEventActionR, &fhir.Reference{
```
reinkrul, line 101: if anyone can access the Questionnaire resource, do we need to audit it? @bramwesselo

orchestrator/careplanservice/handle_getquestionnaireresponse.go
```diff
@@ -75,3 +77,113 @@ func (s *Service) handleReadQuestionnaireResponse(ctx context.Context, request F
...
+		// Try to find a Task that references this QuestionnaireResponse in output
+		taskBundle, err := s.searchTask(ctx, url.Values{"output-reference": []string{"QuestionnaireResponse/" + *qr.Id}}, headers, principal)
+		if err != nil {
+			log.Ctx(ctx).Error().Err(err).Msgf("Error checking tasks for QuestionnaireResponse/%s", *qr.Id)
+			continue
+		}
+
+		// If task is found, the user has access
+		if len(taskBundle.Entry) > 0 {
+			allowedQuestionnaireResponseRefs = append(allowedQuestionnaireResponseRefs, "QuestionnaireResponse/"+*qr.Id)
```
reinkrul, line 170: Duplicate authz code with `handleReadQuestionnaireResponse`, future risk

reinkrul (changes requested):

orchestrator/careplanservice/handle_getcondition.go
dwelman: I think this is worth evaluating and implementing at a higher level for all our CRUD resources, I will add this to the ongoing discussion document of the ORCA improvements we need to make

orchestrator/careplanservice/handle_getcondition.go
dwelman: Yes I think we should definitely implement something like this, it is a common pattern. Will add this to our discussion doc

orchestrator/careplanservice/handle_getquestionnaireresponse.go
dwelman: I will abstract out the auth check and leave a comment that we need to investigate further to do this kind of abstraction across all resource types

reinkrul (approved):

PR #604, opened 2025-05-08
DEV-145 | Implement resource-creator based authz

dwelman:
When a resource is created, append the Extension field with the creating organisation.

This can be used for authz when trying to read or update the resource.

Previously this was attempted with FHIR AuditEvent and subsequently disabled due to unsuitability, this change reimplements it with the new authz code.


orchestrator/careplanservice/authz.go
```diff
@@ -217,40 +220,35 @@ type CreatorPolicy[T any] struct {
...
 
-	return &PolicyDecision{
+	for _, extension := range extensions {
+		if extension.Url == CreatorExtensionURL && extension.ValueReference != nil && extension.ValueReference.Identifier != nil {
+			// Compare with principal's organization identifiers
+			for _, orgIdentifier := range principal.Organization.Identifier {
+				if coolfhir.IdentifierEquals(extension.ValueReference.Identifier, &orgIdentifier) {
+					return &PolicyDecision{
 		Allowed: true,
 		Reasons: []string{"CreatorPolicy: not implemented, anyone has access"},
```
reinkrul, line 246: this reason is now wrong

orchestrator/careplanservice/handle_createtask.go
```diff
@@ -199,6 +218,16 @@ func (s *Service) handleCreateTask(ctx context.Context, request FHIRHandlerReque
 			return nil, coolfhir.NewErrorWithCode("Task.For must be set with a local reference, or a logical identifier, referencing a patient", http.StatusBadRequest)
 		}
 
+		task.Extension = []fhir.Extension{
```
reinkrul, line 221: maybe refactor this to a SetCreator() function?

orchestrator/careplanservice/operation_update_test.go
```diff
@@ -93,6 +93,8 @@ func TestFHIRUpdateOperationHandler_Handle(t *testing.T) {
 				assertBundleEntry(t, tx, coolfhir.EntryIsOfType(*task1Ref.Type), func(t *testing.T, entry fhir.BundleEntry) {
 					assert.Equal(t, fhir.HTTPVerbPUT, entry.Request.Method)
 					assert.Equal(t, *task1Ref.Reference, entry.Request.Url)
+					// Validate resource creation extension
+					updatedTask.Extension = TestCreatorExtension
```
reinkrul, line 97: Do we need to assert, in `operation_update.go` that nobody can update (change) the extension?

reinkrul (changes requested):

orchestrator/careplanservice/authz.go
```diff
@@ -217,40 +203,44 @@ type CreatorPolicy[T any] struct {
 }
 
 func (o CreatorPolicy[T]) HasAccess(ctx context.Context, resource T, principal auth.Principal) (*PolicyDecision, error) {
-	// TODO: Find a more suitable way to handle this auth.
-	// The AuditEvent implementation has proven unsuitable and we are using the AuditEvent for unintended purposes.
-	// For now, we can return true, as this will follow the same logic as was present before implementing the AuditEvent.
+	// Check if the resource has a resource-creator extension matching the principal's organization identifier
+	resourceValue := reflect.ValueOf(resource)
```
dwelman, line 207: Reflection is not the best solution, I am going to investigate modifying the FHIR Go code generator to have a BaseResource type with support to get and set Extension

orchestrator/careplanservice/handle_createtask.go
dwelman: As discussed on Slack, I've implemented a base `fhir.HasExtension` type and we can now call `SetExtension` on our fhir resources directly

orchestrator/careplanservice/operation_update_test.go
dwelman: That is a good idea, will implement

orchestrator/careplanservice/operation_update.go
```diff
@@ -81,6 +82,15 @@ func (h FHIRUpdateOperationHandler[T]) Handle(ctx context.Context, request FHIRH
 			StatusCode: http.StatusForbidden,
 		}
 	}
+
+	// Validate that the resource Extension is not being modified
+	if !equalExtensions(resource.GetExtension(), existingResource.GetExtension()) {
```
reinkrul, line 87: Why not just set it (the appropriate extension)? The caller might add another non-SCP extension (which could be a valid use case). Although it _probably_ won't break for now.

reinkrul (changes requested):

orchestrator/careplanservice/operation_update.go
dwelman: Yeah fair point, I'll validate that the resource-creator is not changed but still allow the changing of other extension fields, that is definitely more future proof

orchestrator/careplanservice/operation_create.go
```diff
@@ -53,6 +53,18 @@ func (h FHIRCreateOperationHandler[T]) Handle(ctx context.Context, request FHIRH
 		resourceBundleEntry.FullUrl = to.Ptr("urn:uuid:" + uuid.NewString())
 	}
 	idx := len(tx.Entry)
+
+	// Set the creator of the resource in the resource.Extension. This is used for creator-based auth
+	resource.SetExtension([]fhir.Extension{
```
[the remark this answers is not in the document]
dwelman: Yeah I can do that, will future proof this

orchestrator/careplanservice/operation_update.go
```diff
@@ -81,6 +81,24 @@ func (h FHIRUpdateOperationHandler[T]) Handle(ctx context.Context, request FHIRH
...
+		}
+	}
+	// Set updated resource creator to that of the existing resource
+	existingResourceExtension := existingResource.GetExtension()
+	for i := range existingResourceExtension {
+		if existingResourceExtension[i].Url == CreatorExtensionURL {
+			extension = append(extension, existingResourceExtension[i])
+			break
+		}
+	}
```
reinkrul, line 99: duplicate code with the create handler, why not have a `SetResourceCreator(resource HasExtension, identifier fhir.Identifier)` function?

reinkrul (approved):
deudplicate code, rest is fine

orchestrator/careplanservice/operation_update.go
dwelman: I have done so, thanks

PR #614, opened 2025-05-12
DEV-212: Introduce new simplified remote FHIR API query proxy endpoint

reinkrul:
Also adds support for querying remote web app endpoints for showing the link in the Frontend.

### DIFF
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -35,7 +35,7 @@ import (
 const basePath = "/cpc"
 
 // scpIdentifierHeaderKey specifies the HTTP request header used to specify the identifier of the external entity (care organization),
-// which's FHIR API should be queried. The identifier is in the form of <system>|<value>.
+// whose FHIR API should be queried. The identifier is in the form of <system>|<value>.
 const scpEntityIdentifierHeaderKey = "X-Scp-Entity-Identifier"
 
 // carePlanURLHeaderKey specifies the HTTP request header used to specify the SCP context, which is a reference to a FHIR CarePlan. Authorization is evaluated according to this CarePlan.


frontend/app/applaunch.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    // if no endpoints found, return undefined
+    if (endpoints.entry?.length == 0) {
+        return undefined
+    }
+    if (endpoints.entry?.length != 1) {
+        // might want to support a list in the future
+        console.warn("More than one launchable application found for organization", organization, endpoints.entry);
+    }
+    return {
```
reinkrul, line 43:
```suggestion
    // Only show first one for now
    return {
```

dwelman (approved):

orchestrator/careplancontributor/service.go
```diff
@@ -274,6 +278,19 @@ func (s *Service) RegisterHandlers(mux *http.ServeMux) {
 	// The section below defines endpoints used for integrating the local EHR with ORCA.
 	// They are NOT specified by SCP. Authorization is specific to the local EHR.
 	//
+	// This endpoint is used by the EHR and ORCA Frontend to query the FHIR API of a remote SCP-node.
+	// The remote SCP-node to query can be specified using the following HTTP headers:
+	// - X-Scp-Entity-Identifier: Uses the identifier of the SCP-node to query (in the form of <system>|<value>), to resolve the registered FHIR base URL
+	mux.HandleFunc(basePath+"/external/fhir/{rest...}", s.withSessionOrBearerToken(func(writer http.ResponseWriter, request *http.Request) {
```
dwelman, line 284: Will `withSessionOrBearerToken` be expanded to support auth via entra/oidc or is that separate work that needs to be done in another area?

reinkrul: I'd expect auth to happen in that middleware, indeed. (probably have to renamed it a bit)

PR #623, opened 2025-05-13
DEV-200: authenticate OIDC clients with secret

reinkrul:


orchestrator/careplancontributor/oidc/storage_test.go
```diff
@@ -76,3 +76,23 @@ func TestStorage_CreateAccessToken(t *testing.T) {
...
 	assert.Equal(t, scopes, storage.tokens[tokenID].Scopes)
 }
+
+func TestStorage_AuthorizeClientIDSecret(t *testing.T) {
+	storage := Storage{
+		mux: &sync.RWMutex{},
+		clients: map[string]Client{
+			"client": {
+				id:     "client",
+				secret: "ba80a943e1c6c6d9939c189df994fb4f2b5fa9c106c49580acbde7f53b087bda", // 'foo', salted with client_id in the form of `client|secret`
```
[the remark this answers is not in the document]
reinkrul: This is a test, so it's intended to be fixed

dwelman (approved):

orchestrator/README.md
```diff
@@ -49,6 +49,9 @@ To configure the OIDC Provider, set the following environment variables:
 To register a client (application), set the following environment variables for that client:
 - `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_ID`: ID of the client, which will be used as `client_id` in the OIDC flow.
 - `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_REDIRECTURI`: URL to which the OIDC provider will redirect the user after authentication.
+- `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_SECRET`: secret of the client to authenticate the client.
```
reinkrul, line 52:
```suggestion
- `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_SECRET`: `client_secret` to authenticate the client.
```

orchestrator/README.md
```diff
@@ -49,6 +49,9 @@ To configure the OIDC Provider, set the following environment variables:
 To register a client (application), set the following environment variables for that client:
 - `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_ID`: ID of the client, which will be used as `client_id` in the OIDC flow.
 - `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_REDIRECTURI`: URL to which the OIDC provider will redirect the user after authentication.
+- `ORCA_CAREPLANCONTRIBUTOR_OIDC_CLIENTS_<clientname>_SECRET`: `client_secret` to authenticate the client.
+  It is the hex-encoded SHA-256 hash of `<client_id>|<secret>`:
```
meelisd, line 53: From what I understand from the code the secret that needs to be specified here should not be the hex encoded hash but it's the client that needs to send it that way. If that is the case this description feels a bit ambiguous, maybe change it to something like this to clarify:

```suggestion
  The client is expected to send the hex-encoded SHA-256 hash of `<client_id>|<secret>` as the secret:
```

reinkrul: No, the client sends its secret in plaintext, and we (IDP) hash it because we configure it already hashed. Reason for this is; at first it was configured in plain text, but I then I realized we can do better (mitigating leaked configuration/compromised admins) by hashing it (just like you don't store passwords plaintext). I added a salt to mitigate rainbow tables.

meelisd (approved):

PR #624, opened 2025-05-13
Remove CPS proxy endpoint in favor of generic SCP proxy endpoint

reinkrul:
We introduced `/cpc/external/fhir`, a new proxy endpoint to which all other outbound FHIR endpoints should migrate. This PR removes the `/cpc/cps/fhir` endpoint, which is used to either call the local CPS (ORCA Frontend, to create Tasks) or the Viewer (to fetch CarePlans).

This removes the `X-Cps-Url` header on the old endpoint, in favor of the new `X-Scp-Fhir-Url` header on the new endpoint. This header can be used to indicate CPC or CPS FHIR APIs, not just CPS.

It can also still be used by the ORCA Frontend to call the local CPS API (like the old endpoint supported), by specifying `local-cps`.

### DIFF
--- a/orchestrator/careplancontributor/service.go
+++ b/orchestrator/careplancontributor/service.go
@@ -38,12 +38,13 @@ const basePath = "/cpc"
 // whose FHIR API should be queried. The identifier is in the form of <system>|<value>.
 const scpEntityIdentifierHeaderKey = "X-Scp-Entity-Identifier"
 
+// scpFHIRBaseURL specifies the HTTP request header used to specify the FHIR base URL, secured through SCP,
+// which should be queried.
+const scpFHIRBaseURL = "X-Scp-Fhir-Url"
+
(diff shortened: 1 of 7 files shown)


orchestrator/careplancontributor/service.go
```diff
@@ -823,16 +780,38 @@ func (s Service) defaultCreateFHIRClientForURL(ctx context.Context, fhirBaseURL
...
+	for _, header := range []string{scpEntityIdentifierHeaderKey, scpFHIRBaseURL} {
 		headerValue := request.Header.Get(header)
 		if headerValue == "" {
 			continue
 		}
 		switch header {
+		case scpFHIRBaseURL:
+			var err error
+			if headerValue == "local-cps" {
+				// TODO: This should probably be optimized by the EHR/ORCA Frontend having a internal endpoint on the CPS, instead.
```
dwelman, line 798: Good callout, but agreed we can hold off on this until later. Let's get this new endpoint working first

dwelman (approved):

PR #635, opened 2025-05-19
DEV-19 | Implement azure adb2c token client

dwelman:
Adds a client that can parse an ADB2C JWT, validate it is not expired, validate it contains required claims, and compares it against a list of known issuers to ensure the token comes from a source we trust.

Automatically handles key rotation for discovery endpoints.

This will be added to CPC in a future commit, for now this is a self-contained change, with extensive unit tests and an integration test that can be used with a real JWT and discovery endpoint


jdksloan (approved):

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,401 @@
...
+func (c *Client) refreshConfigurationsIfNeeded(ctx context.Context) error {
+	if time.Since(c.lastRefresh) < c.refreshInterval {
+		return nil
+	}
+
+	return c.refreshAllConfigurations(ctx)
+}
+
+// ParseToken parses a JWT token without validation and returns the claims
+func ParseToken(tokenString string) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
dwelman: yes, will do

orchestrator/lib/az/adb2c/integration_test.go
```diff
@@ -0,0 +1,122 @@
...
+		case []interface{}:
+			if len(aud) > 0 {
+				clientID = aud[0].(string)
+			}
+		}
+		t.Logf("Extracted client ID from token: %s", clientID)
+	}
+
+	// Get trusted issuers configuration
+	trustedIssuersRaw := os.Getenv("ADB2C_TRUSTED_ISSUERS")
```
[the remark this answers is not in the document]
dwelman: I looked into using koanf but it doesn't play nicely with URL values as keys, so either we need to process a key like HTTP_EXAMPLE_COM_V2_KEYS into a real URL or we process a list into a map. Considering this is just for a test I'll leave it as it is, we can decide how to load the issuers into the client when this is implemented in CPC

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := validatedClaims.ValidateIssuedAt(); err != nil {
+		return nil, err
+	}
+
+	return validatedClaims, nil
+}
+
+// getPublicKeyFromJWKS fetches the public key corresponding to the key ID from the JWKS endpoint
+func (c *Client) getPublicKeyFromJWKS(ctx context.Context, jwksURI, kid, issuer string) (*rsa.PublicKey, error) {
+	jwks, err := c.jwksFetcher.Fetch(ctx, jwksURI)
```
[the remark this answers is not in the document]
dwelman: The current implementation does cache keys and will perform a refresh after 24h when keys are requested

jdksloan (approved):

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,177 @@
...
+	Enabled bool `koanf:"enabled"`
+	// ADB2CClientID is the name of the ADB2C client to use for authentication.
+	ADB2CClientID string `koanf:"clientid"`
+	// ADB2CTrustedIssuers is a map of friendly names to trusted issuer configurations.
+	// The friendly names are used as environment variable suffixes.
+	ADB2CTrustedIssuers map[string]TrustedIssuer `koanf:"trustedissuers"`
+}
+
+// ToTrustedIssuersMap converts the config format to the format expected by the ADB2C client
+func (c Config) ToTrustedIssuersMap() map[string]string {
```
reinkrul, line 32: nitpick: I think you can drop `To`

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,177 @@
...
+		// Additional security checks for discovery URL
+		if discoveryURL.Host == "" {
+			return fmt.Errorf("discovery URL must have a valid host for issuer '%s'", name)
+		}
+
+		// Prevent localhost/private IPs in production (optional - can be configured)
+		if isPrivateOrLocalhost(issuerURL.Host) || isPrivateOrLocalhost(discoveryURL.Host) {
+			// This is a warning-level check - you might want to make this configurable
+			// return fmt.Errorf("private/localhost URLs not allowed in production for issuer '%s'", name)
+		}
```
reinkrul, line 101: you could pass in `strictMode bool` to this function, and have this return an error in strict mode

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,177 @@
...
+// DefaultConfig returns the default ADB2C configuration.
+func DefaultConfig() Config {
+	return Config{
+		Enabled:             false,
+		ADB2CClientID:       "",
+		ADB2CTrustedIssuers: make(map[string]TrustedIssuer),
+	}
+}
+
+func loadConfigInto(target any) error {
```
reinkrul, line 147: why do we need custom config loading for this? Wouldn't you be able to just use whatever map key (that's what we do for the OIDC config), e.g.:

```
ORCA_ADB2C_TRUSTEDISSUERS_0_ISSUERURL=XYZ
ORCA_ADB2C_TRUSTEDISSUERS_0_DISCOVERYURL=XYZ
```
?

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// LoadConfig loads the ADB2C configuration from environment variables using koanf.
+// Environment variables should be prefixed with ADB2C_ and use underscores for nested keys.
+// For example:
+//   - ADB2C_ENABLED=true
+//   - ADB2C_CLIENTID=my-client-id
+//   - ADB2C_TRUSTEDISSUERS_TENANT1_ISSUERURL=https://tenant1.b2clogin.com/tenant1.onmicrosoft.com/v2.0/
+//   - ADB2C_TRUSTEDISSUERS_TENANT1_DISCOVERYURL=https://tenant1.b2clogin.com/tenant1.onmicrosoft.com/v2.0/.well-known/openid_configuration
+func LoadConfig() (*Config, error) {
```
reinkrul, line 107: still there?

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+	directJwksUris := true
+	for _, endpoint := range trustedIssuers {
+		if !strings.HasSuffix(endpoint, "/keys") &&
+			!strings.HasSuffix(endpoint, "/jwks") &&
+			!strings.HasSuffix(endpoint, "/jwks.json") &&
+			!strings.Contains(endpoint, "/keys?p=") {
+			directJwksUris = false
+			break
+		}
+	}
```
reinkrul, line 119: this looks a bit hacky, or is this per RFC/Azure feature?

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+		config, err := c.fetchOpenIDConfigurationFromURL(ctx, discoveryURL)
+		if err != nil {
+			return fmt.Errorf("failed to fetch OpenID configuration for issuer %s: %w", issuer, err)
+		}
+
+		// Verify the issuer in the config matches the expected issuer
+		// Skip this check for direct JWKS URIs (already handled above)
+		if config.Issuer != issuer && config.Issuer != "" {
+			return fmt.Errorf("issuer mismatch: expected %s, got %s", issuer, config.Issuer)
+		}
```
reinkrul, line 190: do we want startup to fail if the OpenID Provider has a hickup? You could also lazy-initialize this when a token is checked. Or both: do it on startup (and log when it fails), _and_ when a token is to be checked (and this particular provider isn't initialized).

It's an edge case though, so feel free to ignore or do it at some later point.

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+		return nil, fmt.Errorf("failed to fetch OpenID configuration: status code %d", resp.StatusCode)
+	}
+
+	// Limit response size to prevent DoS attacks
+	limitedReader := &io.LimitedReader{R: resp.Body, N: 1024 * 1024} // 1MB limit
+
+	var config oidc.DiscoveryConfiguration
+	if err := json.NewDecoder(limitedReader).Decode(&config); err != nil {
+		return nil, fmt.Errorf("failed to parse OpenID configuration: %w", err)
+	}
```
reinkrul, line 258: good

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+	if !ok {
+		return nil, fmt.Errorf("no configuration found for issuer: %s", issuer)
+	}
+
+	// Check if refresh is needed
+	if time.Since(c.lastRefresh) >= c.refreshInterval {
+		// Since we do not store the last refresh time per issuer/token, we refresh all configurations
+		if err := c.refreshAllConfigurations(ctx); err != nil {
+			return nil, fmt.Errorf("failed to refresh configurations: %w", err)
+		}
```
reinkrul, line 278: do we expect to need multiple trusted issuers? Since we're adding risk here (one openid configuraiton expired, need to refresh all, the second one is down blocking all access). I'd say we support it properly, or just support 1?

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,134 @@
+package adb2c
```
reinkrul, line 1: I don't think this is even adb2c-specific now?

orchestrator/lib/az/adb2c/config.go
```diff
@@ -0,0 +1,134 @@
...
+	// IssuerURL is the issuer URL from the JWT token (e.g., https://tenant.b2clogin.com/tenant.onmicrosoft.com/v2.0/)
+	IssuerURL string `koanf:"issuerurl"`
+	// DiscoveryURL is the OpenID Connect discovery endpoint or direct JWKS URL
+	DiscoveryURL string `koanf:"discoveryurl"`
+}
+
+type Config struct {
+	Enabled bool `koanf:"enabled"`
+	// ADB2CClientID is the name of the ADB2C client to use for authentication.
+	ADB2CClientID string `koanf:"clientid"`
```
reinkrul, line 24: this is not used? Shouldn't we check the `aud` claim with this?

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	for iter := keySet.Keys(ctx); iter.Next(ctx); {
+		pair := iter.Pair()
+		key := pair.Value.(jwk.Key)
+
+		if key.Algorithm() == jwa.InvalidKeyAlgorithm("") {
+			// The 'alg' field is optional for Azure ADB2C keys, but required by jwk, so we set a default
+			// For RSA and EC key types, RS256 is a reasonable default
+			if key.KeyType() == jwa.RSA || key.KeyType() == jwa.EC {
```
reinkrul, line 528: `RS256` = `PKCSv1.5` signatures, which is RSA. EC (elliptic curve) is an entirely different cryptographic suite, with other (non-compatible) algotighms (e.g. `P256`). So setting RS256 makes no sense if the key type is EC; https://www.sectigo.com/resource-library/rsa-vs-dsa-vs-ecc-encryption

RSA uses a modulo for it's cryptography, EC a sort of billard ball bouncing on a curve (the 2D curve being the billard ball table).

orchestrator/lib/az/adb2c/tokens_test.go
```diff
@@ -0,0 +1,713 @@
...
+		"exp": time.Now().Add(time.Hour).Unix(),
+		"iat": time.Now().Add(-time.Minute).Unix(),
+	}, "invalid-key-id")
+
+	tests := []struct {
+		name          string
+		token         string
+		expectedError error
+	}{
+		{
```
reinkrul, line 424: tests for;
- replayed token
- invalid signature ('crafted' JWT)

orchestrator/lib/az/adb2c/tokens.go
```diff
@@ -0,0 +1,533 @@
...
+				key.Set(jwk.AlgorithmKey, jwa.RS256)
```
reinkrul, line 529: doesn't azure use `PS256` (which is safer)? weird... 

PS256 (PKCSv1.5 PSS) is like RS256, but uses a "masked salting function" which is recommended, nowadays (don't ask me to explain it)

reinkrul (changes requested):

orchestrator/lib/az/adb2c/tokens.go
dwelman: Initially when I was having trouble getting the key to work I was experimenting with different endpoint configurations, this is not actually used or needed so I will remove it

orchestrator/lib/az/adb2c/tokens.go
dwelman: I have a surface level understanding of cryptography, let me rather remove EC since we are only using RSA anyway. Thanks

orchestrator/lib/az/adb2c/tokens.go
dwelman: Noted, it seems to work fine with PS256 so we can use that instead

orchestrator/lib/az/adb2c/tokens.go
dwelman: I think it should be simple enough to add a refresh per issuer, and perform lazy loading. This is quite fragile regardless and we don't want CPC crashing if it can't reach the key provider

reinkrul (approved):

PR #637, opened 2025-05-21
DEV-249: Support multiple client certificates for outbound requests

reinkrul:


reinkrul:
Let's merge this after the tests tomorrow, just in case

PR #639, opened 2025-05-22
DEV-255 | Reload page if network issues cause task issues

dwelman:


frontend/app/enrollment/task/[taskId]/page.tsx
```diff
@@ -53,6 +54,16 @@ export default function EnrollmentTaskPage() {
             })
     }, [serviceRequest])
 
+    useEffect(() => {
+        if (shouldReload) {
+            const timeout = setTimeout(() => {
+                window.location.reload();
+                setShouldReload(false);
+            }, 10000);
```
Comment (reinkrul) [inline frontend/app/enrollment/task/[taskId]/page.tsx:62]: it's expected to be a race condition which is resolved very, very fast, so I'd reload after 1 second

Comment (dwelman) [reply frontend/app/enrollment/task/[taskId]/page.tsx:62]: oops, too many 0s

frontend/app/enrollment/task/[taskId]/page.tsx
```diff
@@ -53,6 +54,16 @@ export default function EnrollmentTaskPage() {
             })
     }, [serviceRequest])
 
+    useEffect(() => {
+        if (shouldReload) {
+            const timeout = setTimeout(() => {
+                window.location.reload();
+                setShouldReload(false);
```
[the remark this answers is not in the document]
Comment (dwelman) [reply frontend/app/enrollment/task/[taskId]/page.tsx:61]: yes, as it will enter the condition setting shouldReload to true again. I don't know how valuable this stopgap is when it can be fixed in demo by refreshing the page, maybe worth putting a bit more time and thought into a proper fix for the future?

PR #648, opened 2025-05-28
DEV-20 | Add JWT authentication to CPC

dwelman:


orchestrator/README.md
```diff
@@ -28,6 +28,26 @@ Use the following environment variables to configure the orchestrator:
 - `ORCA_CAREPLANCONTRIBUTOR_FRONTEND_URL`: Base URL of the frontend application, to which the browser is redirected on app launch (default: `/frontend/enrollment`).
 - `ORCA_CAREPLANCONTRIBUTOR_SESSIONTIMEOUT`: Configure the user session timeout, use Golang time.Duration format (default: 15m).
 
+### Token Client configuration
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_ENABLED`: Enable the Token Client (default: `false`).
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_CLIENTID`: ClientID used for token validation, this will typically be the same as the `aud` claim in the JWT being validated.
+The following two fields can be repeated for multiple trusted issuers, but must both be set for each issuer:
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_<ISSUER_NAME>_ISSUERURL`: Same as the `iss` claim in the JWT.
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_<ISSUER_NAME>_DISCOVERYURL`: Same as the `iss` claim in the JWT, but with `tfp` before `/v2.0/` as well as the `.well-known/openid_configuration` suffix.
```
reinkrul, line 36: it seems `tfp` is Microsoft-specific. Maybe good to note in this documentation that it has been tested with MS, but should work with other OpenID COnnect Providers?

orchestrator/README.md
```diff
@@ -28,6 +28,26 @@ Use the following environment variables to configure the orchestrator:
...
+### Token Client configuration
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_ENABLED`: Enable the Token Client (default: `false`).
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_CLIENTID`: ClientID used for token validation, this will typically be the same as the `aud` claim in the JWT being validated.
+The following two fields can be repeated for multiple trusted issuers, but must both be set for each issuer:
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_TRUSTEDISSUERS_<ISSUER_NAME>_ISSUERURL`: Same as the `iss` claim in the JWT.
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_TRUSTEDISSUERS_<ISSUER_NAME>_DISCOVERYURL`: Same as the `iss` claim in the JWT, but with `tfp` before `/v2.0/` as well as the `.well-known/openid_configuration` suffix.
+Example:
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_TRUSTEDISSUERS_EXAMPLE_ISSUERURL`: `https://your-tenant.b2clogin.com/your-tenant.onmicrosoft.com/v2.0/`
+- `ORCA_CAREPLANCONTRIBUTOR_TOKENCLIENT_TRUSTEDISSUERS_EXAMPLE_DISCOVERYURL`: `https://your-tenant.b2clogin.com/your-tenant.onmicrosoft.com/B2C_1_local_login/v2.0/.well-known/openid_configuration`
+
```
reinkrul, line 50: duplicate?

orchestrator/README.md
```diff
@@ -28,6 +28,26 @@ Use the following environment variables to configure the orchestrator:
 - `ORCA_CAREPLANCONTRIBUTOR_FRONTEND_URL`: Base URL of the frontend application, to which the browser is redirected on app launch (default: `/frontend/enrollment`).
 - `ORCA_CAREPLANCONTRIBUTOR_SESSIONTIMEOUT`: Configure the user session timeout, use Golang time.Duration format (default: 15m).
 
+### Token Client configuration
```
reinkrul, line 31: this is missing some docs on what the feature is for? Maybe a section on API auth (not the Nuts-APIs, but EHR-facing) that describes the options?

orchestrator/careplancontributor/config.go
```diff
@@ -25,6 +26,7 @@ type Config struct {
 	FrontendConfig FrontendConfig   `koanf:"frontend"`
 	AppLaunch      applaunch.Config `koanf:"applaunch"`
 	OIDCProvider   oidc.Config      `koanf:"oidc"`
+	TokenClient    token.Config     `koanf:"tokenclient"`
```
reinkrul, line 29: this actually has some overlap with `oidc`, which is the provider... I wonder if we should rename/restructure things, since `tokenclient` is actually an auth-alternative to `StaticBearerToken`... What do you think?

orchestrator/careplancontributor/service.go
```diff
@@ -524,12 +533,41 @@ func (s Service) handleGetContext(response http.ResponseWriter, _ *http.Request,
 
 func (s Service) withSessionOrBearerToken(next func(response http.ResponseWriter, request *http.Request)) http.HandlerFunc {
```
reinkrul, line 534: can you rename this to something that resembles its intended consumers better? Something maybe `withUserAuth` (also ugly, but might be clearer? `user` also being an API client..)

orchestrator/careplancontributor/service.go
```diff
@@ -524,12 +533,41 @@ func (s Service) handleGetContext(response http.ResponseWriter, _ *http.Request,
...
-			s.SessionManager.Get(request) != nil {
+		// Session will be present for FE requests
+		if s.SessionManager.Get(request) != nil {
+			next(response, request)
+			return
+		}
+
+		bearer := request.Header.Get("Authorization")
+		// Static bearer token, not valid in prod
+		if globals.StrictMode == false && s.config.StaticBearerToken != "" && bearer == "Bearer "+s.config.StaticBearerToken {
```
reinkrul, line 544: maybe we should move the `strictmode && staticbearertoken != ""` check to the config validation..

orchestrator/careplancontributor/service.go
```diff
@@ -524,12 +533,41 @@ func (s Service) handleGetContext(response http.ResponseWriter, _ *http.Request,
...
 			next(response, request)
 			return
 		}
+
+		// Try to validate bearer token as adb2c jwt
+		if s.tokenClient != nil {
+			// Validate the token
+			token := strings.TrimPrefix(bearer, "Bearer ")
+			if token == "" {
+				http.Error(response, "no bearer token found", http.StatusUnauthorized)
```
reinkrul, line 554: I'd not return an error here, and rely on the error on line 571, since it could also mean there wasn't any credential provided (but it was a browser without a cookie).

reinkrul (changes requested):

orchestrator/README.md
dwelman: Weird, I wonder if this was a merge bug or something. Removed anyway

orchestrator/careplancontributor/config.go
dwelman: Good points, let's discuss

orchestrator/careplancontributor/service.go
dwelman: Yeah good idea

orchestrator/careplancontributor/service.go
dwelman: Right, fair enough

orchestrator/README.md
dwelman: Sounds good, will add

PR #681, opened 2025-06-12
Dev 270 orca api validation

jdksloan:


deployments/dev/clinic/data/tunnel.id
dwelman, line 1: We shouldn't be committing these devtunnel files, maybe you can add them to the gitignore?

deployments/dev/hospital/data/tunnel.id
dwelman, line 1: Same as my other comment relating to the clinic tunnel files

deployments/dev/shared_config/discovery/homemonitoring.json
```diff
@@ -0,0 +1,117 @@
+{
+  "id": "dev:HomeMonitoring2024",
+  "endpoint": "https://zh89ms3x-9080.euw.devtunnels.ms/discovery/dev:HomeMonitoring2024",
```
dwelman, line 3: Is this endpoint not going to change each time a new devtunnel is launched? I think this will need to be configurable (bonus points if we can have it auto-configured based on the current devtunnel session)

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -221,17 +222,53 @@ func (s *Service) createSubTaskOrAcceptPrimaryTask(ctx context.Context, cpsClien
...
 						// TODO: why return an error here, and not for the rest?
+						// TODO: Retryable errors shouldn't be TaskRejections
 						return &TaskRejection{
 							Reason:       "Failed to fetch questionnaire",
 							ReasonDetail: err,
 						}
 					}
+
+					// Check if there's a QuestionnaireResponse, and if so, if it's valid
+					if len(task.Output) >= inputIdx {
```
dwelman, line 239: In addition to what Copilot has suggested, I'd recommend something like:
```
if inputIdx >= len(task.Output) {
    continue
}
```

This early return is more idiomatic Go and also means less nesting so the code is easier to read

orchestrator/careplancontributor/taskengine/validation.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	for _, issue := range bundle.Issue {
+		if issue.Severity == fhir.IssueSeverityFatal || issue.Severity == fhir.IssueSeverityError {
+			return &bundle, errors.New("validation failed, see OperationOutcome for details")
+		}
+	}
+	return &bundle, nil
+}
+
+func HasValidationFailed(ctx context.Context, err error) error {
```
dwelman, line 105: Is this actually used anywhere? Just a few lines up a validation failed error is created. Can this be removed?

orchestrator/careplancontributor/taskengine/validation_test.go
```diff
@@ -0,0 +1,88 @@
...
+		require.Error(t, err)
+		require.NotNil(t, ooc)
+		oocJSON, _ := json.Marshal(ooc)
+		require.Contains(t, string(oocJSON), "Item has answer, even though it is not enabled (item id = '170292e5-3163-43b4-88af-affb3e4c27ab')")
+	})
+}
+
+func downloadFHIRValidatorCLI() error {
+	validatorJar := "bin/fhir_validator_cli-6.5.7.jar"
+	// If not exists, download it from https://github.com/hapifhir/org.hl7.fhir.core/releases/download/6.5.7/validator_cli.jar
```
dwelman, line 66: This feels a bit risky. How is the download of this jar file handled in prod?

deployments/.gitignore
```diff
@@ -1,2 +1,6 @@
 *.private.pem
-extra.env
\ No newline at end of file
+extra.env
```
reinkrul, line 2: I don't think you intended to check this in

orchestrator/careplancontributor/handle_taskfiller.go
```diff
@@ -221,17 +222,53 @@ func (s *Service) createSubTaskOrAcceptPrimaryTask(ctx context.Context, cpsClien
...
+							return &TaskRejection{
+								Reason: fmt.Sprintf("Subtask output %d was expected to be a reference to a QuestionnaireResponse", inputIdx),
+							}
+						}
+						questionnaireResponseRef := task.Output[inputIdx].ValueReference.Reference
+						var questionnaireResponse fhir.QuestionnaireResponse
+						err := cpsClient.ReadWithContext(ctx, *questionnaireResponseRef, &questionnaireResponse)
+						if err != nil {
+							// TODO: Retryable errors shouldn't be TaskRejections
+							return &TaskRejection{
```
reinkrul, line 254: I think this can be a normal error, rather than a `TaskRejection`. The latter causes the task status to be set to `rejected`, while it can be a network error/hickup.

orchestrator/careplancontributor/taskengine/validation_test.go
```diff
@@ -0,0 +1,88 @@
...
+func TestFHIRValidatorCLIQuestionnaireResponseValidator_Validate(t *testing.T) {
+	_, err := exec.LookPath("java")
+	if err != nil {
+		t.Skip("not running FHIR validator test, no Java installed")
+	}
+	err = downloadFHIRValidatorCLI()
+	require.NoError(t, err)
+	ctx := context.Background()
+
+	validator := FHIRValidatorAPIQuestionnaireResponseValidator{URL: "http://localhost:3500/validate", client: &http.Client{}}
```
reinkrul, line 25: the JAR is never started, maybe change this test to use testcontainers. Search for `SetupHAPI` in the codebase, for an example with HAPI.

orchestrator/careplancontributor/service.go
```diff
@@ -169,16 +169,17 @@ type Service struct {
...
-	sseService                    *sse.Service
-	createFHIRClientForURL        func(ctx context.Context, fhirBaseURL *url.URL) (fhirclient.Client, *http.Client, error)
-	oidcProvider                  *op.Service
-	httpHandler                   http.Handler
-	tokenClient                   *rp.Client
+	transport                      http.RoundTripper
+	workflows                      taskengine.WorkflowProvider
+	healthdataviewEndpointEnabled  bool
+	notifier                       ehr.Notifier
+	questionnaireResponseValidator taskengine.QuestionnaireResponseValidator
```
reinkrul, line 176: this should probably be enabled when some specific config is set? (e.g. validation endpoint)

reinkrul (changes requested):

deployments/.gitignore
```diff
@@ -1,2 +1,5 @@
 *.private.pem
-extra.env
\ No newline at end of file
+/dev/shared_config/discovery/homemonitoring.json
+/dev/clinic/**/tunnel*
+/dev/hospital/**/tunnel*
```
reinkrul, line 4: it doesn't use tunnels anymore

PR #692, opened 2025-06-22
DEV-406: Support SMART on FHIR as AppLaunch method

reinkrul:
This adds:
- Support for SMART on FHIR applaunch with public client, tested with SoF sandbox (https://launch.smarthealthit.org/)
- Task overview: for SMART on FHIR launches, we're missing the ordered Task/ServiceRequest (we might add that in the future, but not available right now), so to be able to show anything, I added a (very) simple SCP Task overview.

Changes:
- Set cookie path to domain root. Otherwise, the redirect to the SoF OIDC Provider causes the ORCA user session cookie to be lost
- Support app launches without ServiceRequest and/or PractitionerRole

Out of scope:
- Integration with Epic (on FHIR)

Todo:
- [x] Figure out the right cookie settings for zitadel/oidc's `state` cookie, which is lost when the app is opened in an iFrame (although this might not be an issue?)
- [x] Tests
- [x] Documentation
- [x] Support confidential client mode with JWTs


reinkrul:
sof-1
sof-2
sof-3

orchestrator/careplancontributor/applaunch/smartonfhir/service.go
```diff
@@ -2,166 +2,253 @@ package smartonfhir
...
+	cookieEncryptKey := make([]byte, 32)
+	if _, err := rand.Read(cookieEncryptKey); err != nil {
+		// Can't happen, but just in case (we don't want to end up with a zeroed key)
+		panic(err)
+	}
+	var cookieHandlerOpts []zitadelHTTP.CookieHandlerOpt
+	if !strictMode {
+		cookieHandlerOpts = append(cookieHandlerOpts, zitadelHTTP.WithUnsecure())
+	}
+	//cookieHandlerOpts = append(cookieHandlerOpts, zitadelHTTP.WithSameSite(http.SameSiteNoneMode))
```
dwelman, line 89: remove? Or otherwise maybe a comment explaining when and why to uncomment this if it's left in for easier issue resolution in the future?

dwelman (approved):

PR #731, opened 2025-07-25
Fixed the dutch phone number validator

jdksloan:
Updated the dutch phone number validator to be more robust

### DIFF
--- a/orchestrator/careplanservice/validate_patient.go
+++ b/orchestrator/careplanservice/validate_patient.go
@@ -80,8 +80,26 @@ func validatePhone(phone *string) error {
 		return errors.New("phone number is required")
 	}
 
-	if !strings.HasPrefix(*phone, "+31") {
-		return errors.New("phone number must start with +31")
+	return isDutchPhoneNumber(phone)
+}
+
+func isDutchPhoneNumber(phone *string) error {
+	normalised := strings.NewReplacer("-", "", " ", "", "(", "", ")", "").Replace(strings.TrimSpace(*phone))
+
+	normalised = strings.Map(func(r rune) rune {
+		if r >= '0' && r <= '9' || r == '+' {
+			return r
+		}
+		return -1
+	}, normalised)
+
+	if len(normalised) == 10 && strings.HasPrefix(normalised, "06") {
+		return nil
 	}
-	return nil
+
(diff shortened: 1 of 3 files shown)


orchestrator/careplanservice/validate_patient.go
```diff
@@ -80,8 +80,26 @@ func validatePhone(phone *string) error {
...
 
-	if !strings.HasPrefix(*phone, "+31") {
-		return errors.New("phone number must start with +31")
+	return isDutchPhoneNumber(phone)
+}
+
+func isDutchPhoneNumber(phone *string) error {
+	normalised := strings.NewReplacer("-", "", " ", "", "(", "", ")", "").Replace(strings.TrimSpace(*phone))
+
+	normalised = strings.Map(func(r rune) rune {
```
reinkrul, line 89: looks like you're doing the same normalization twice now?, the map covers every case the replacer covers?

orchestrator/careplanservice/validate_patient.go
```diff
@@ -80,8 +80,26 @@ func validatePhone(phone *string) error {
...
+		}
+		return -1
+	}, normalised)
+
+	if len(normalised) == 10 && strings.HasPrefix(normalised, "06") {
+		return nil
 	}
-	return nil
+
+	if len(normalised) == 12 && strings.HasPrefix(normalised, "+316") {
```
reinkrul, line 100: 00316... isn't a valid case? (+ is short hand for 00)

reinkrul (changes requested):

orchestrator/careplanservice/validate_patient.go
```diff
@@ -80,8 +80,26 @@ func validatePhone(phone *string) error {
...
+
+func isDutchPhoneNumber(phone *string) error {
+	normalised := strings.NewReplacer("-", "", " ", "", "(", "", ")", "").Replace(strings.TrimSpace(*phone))
+
+	normalised = strings.Map(func(r rune) rune {
+		if (r >= '0' && r <= '9') || r == '+' {
+			return r
+		}
+		return -1
+	}, normalised)
```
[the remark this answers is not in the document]
reinkrul: I like the 'map' better than the regex

reinkrul (approved):

PR #737, opened 2025-07-27
DEV-588: Make CPC FHIR API configuration multi-tenant

reinkrul:
This PR builds on https://github.com/SanteonNL/orca/pull/732


orchestrator/careplancontributor/applaunch/zorgplatform/service.go
```diff
@@ -203,53 +197,71 @@ type Service struct {
...
-	accessTokenCache *ttlcache.Cache[string, string]
 }
 
 func (s *Service) RegisterHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("POST "+appLaunchUrl, s.handleLaunch)
 }
 
-func (s *Service) EhrFhirProxy() (coolfhir.HttpProxy, fhirclient.Client) {
+// CreateEHRProxies creates HTTP proxies and FHIR clients to interact with the Zorgplatform FHIR API.
+// It should only be invoked once, because it creates caches with background goroutines.
```
dwelman, line 207: Maybe it is worth tracking how many times we invoke this and returning early if it has already been called? Kind of a hacky lazy singleton?

dwelman (approved):

orchestrator/careplancontributor/applaunch/zorgplatform/service.go
reinkrul: It's only called by `cmd/server.go` on initialization, so really not worth it imo

PR #739, opened 2025-07-28
DEV-588: Validate tenant ID

reinkrul:
If it contains characters not allowed in URL paths, things will get ugly.

### DIFF
--- a/orchestrator/cmd/tenants/tenant.go
+++ b/orchestrator/cmd/tenants/tenant.go
@@ -13,6 +13,9 @@ type Config map[string]Properties
 
 func (c Config) Validate(cpsEnabled bool) error {
 	for id, props := range c {
+		if !isIDValid(id) {
+			return fmt.Errorf("tenant %s: invalid ID", id)
+		}
 		if props.Nuts.Subject == "" {
 			return fmt.Errorf("tenant %s: missing Nuts subject", id)
 		}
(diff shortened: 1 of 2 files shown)


orchestrator/cmd/tenants/tenant.go
```diff
@@ -30,6 +33,16 @@ func (c Config) Validate(zorgplatformEnabled bool) error {
 	return nil
 }
 
+func isIDValid(tenantID string) bool {
```
dwelman, line 36: Couldn't we do this with regex? Might be a bit cleaner and easier to modify in the future

orchestrator/cmd/tenants/tenant.go
```diff
@@ -30,6 +33,16 @@ func (c Config) Validate(zorgplatformEnabled bool) error {
...
+	// Only alphanumeric, dashes, and underscores are allowed in tenant IDs
+	for _, r := range tenantID {
+		if !(r >= '0' && r <= '9' || r >= 'a' && r <= 'z' || r >= 'A' && r <= 'Z' || r == '-' || r == '_') {
+			return false
+		}
+	}
+	return len(tenantID) > 0
```
dwelman, line 43: Does a tenant ID have a min length? Would "A" or "1" be a valid tenant ID?

orchestrator/cmd/tenants/tenant.go
reinkrul: We could, although regex are much, much slower. Although this is only evaluated once on startup, so that's really not a valid argument. I think it's potato/potato. If you prefer a regex, I can change it.

PR #759, opened 2025-08-05
[SPIKE] Implement OTEL tracing in ORCA

dwelman:
Adds basic OTEL tracing to the core ORCA functionality, with basic testing and confirmation of export in E2E tests


zidad (approved):

reinkrul (changes requested):
If spans always specify the name of the function they're executing in, it might be worth considering looking it up using the `runtime` package; https://gist.github.com/HouLinwei/16df41bee7d799f0928e717b23d97a9b

orchestrator/careplancontributor/taskengine/workflow.go
```diff
@@ -4,16 +4,29 @@ import (
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var ErrWorkflowNotFound = errors.New("workflow not found")
 
+// stringValue safely extracts string value from pointer, returning empty string if nil
+func stringValue(s *string) string {
```
reinkrul, line 23: I think we have something similar in our `to` package

orchestrator/careplanservice/operation_search.go
```diff
@@ -24,10 +27,36 @@ type FHIRSearchOperationHandler[T any] struct {
...
+		attribute.Int("fhir.search.param_count", len(request.QueryParams)),
+	)
+
+	// Add search parameters as span attributes for better observability
+	searchParams := make([]string, 0, len(request.QueryParams))
+	for param, values := range request.QueryParams {
+		searchParams = append(searchParams, param+"="+strings.Join(values, ","))
+	}
+	if len(searchParams) > 0 {
+		span.SetAttributes(attribute.StringSlice("fhir.search.parameters", searchParams))
```
reinkrul, line 52: We risk logging BSNs here, so maybe we don't want to do this?

reinkrul:
would it make sense to have constants (somewhere central) for trace attributes that are used in more than 1 (or 2) instances? Then we improve homogentiy of the attributes, which I suppose improves searchabiility of the traces (if attributes are names the same across differing spans)

orchestrator/lib/otel/tracing_middleware.go
```diff
@@ -0,0 +1,53 @@
...
+func (rw *responseWriter) WriteHeader(code int) {
+	rw.statusCode = code
+	rw.ResponseWriter.WriteHeader(code)
+}
+
+func getTenantFromRequest(r *http.Request) string {
+	if tenant := r.PathValue("tenant"); tenant != "" {
+		return tenant
+	}
+	return "unknown"
```
reinkrul, line 52: Better to not set the attribute at all? Because some endpoints are not tenant-specific.

orchestrator/lib/otel/config.go
```diff
@@ -0,0 +1,271 @@
...
+		for _, pair := range pairs {
+			if kv := strings.SplitN(strings.TrimSpace(pair), "=", 2); len(kv) == 2 {
+				resourceAttributes[strings.TrimSpace(kv[0])] = strings.TrimSpace(kv[1])
+			}
+		}
+	}
+
+	// Log the effective configuration
+	// TODO: remove after debugging
+	fmt.Printf("Using OTLP endpoint: %s (protocol: %s, insecure: %t)\n", endpoint, protocol, insecure)
```
reinkrul, line 112: this

orchestrator/cmd/server.go
```diff
@@ -30,6 +31,28 @@ func Start(ctx context.Context, config Config) error {
...
+	}
+
+	// Ensure proper cleanup of OpenTelemetry on shutdown
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := tracerProvider.Shutdown(shutdownCtx); err != nil {
+			log.Error().Err(err).Msg("Failed to shutdown OpenTelemetry")
+		} else {
+			log.Info().Msg("OpenTelemetry shutdown successfully")
```
reinkrul, line 52: debug

orchestrator/careplancontributor/taskengine/workflow.go
dwelman: Oops, good point. Thanks

orchestrator/careplanservice/operation_search.go
dwelman: Good point. I don't think we need this as a trace, I will remove it

reinkrul (approved):

PR #793, opened 2025-08-25
DEV-764: Data import feature

reinkrul:
This PR adds endpoints to import data, which can be used to migrate tasks/service requests made outside of ORCA. It introduces 2 operations:
- `/cpc/<tenant>/fhir/$import`: task details, looks up the patient and creates the Patient, Task, ServiceRequest and CarePlan resources. It then calls the CPS's `$import` operation with those resources in a transaction bundle. The task requester will be the local tenant, the task owner will be the party that invoked the import operation.
- `/cps/<tenant>/$import`: takes a FHIR bundle and directly executed it on the CPS's backing FHIR store. It can only contain resource creates (POST), and only be invoked by the local tenant.

### DIFF
--- a/deployments/dev/docker-compose.yaml
+++ b/deployments/dev/docker-compose.yaml
@@ -1,4 +1,16 @@
 services:
+  aspire-dashboard:
+    image: mcr.microsoft.com/dotnet/aspire-dashboard
+    ports:
+      - "18888:18888"  # Dashboard UI
+      - "18889:18889"  # OTLP gRPC endpoint
+    environment:
+      - DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS=true
+    volumes:
+      - ./aspire-dashboard-data:/data
+    restart: unless-stopped
+    networks:
+      - orca-shared-services
   nutsnode:
     image: nutsfoundation/nuts-node:6.1.2
     ports:
@@ -15,6 +27,8 @@ services:
       - NUTS_DISCOVERY_DEFINITIONS_DIRECTORY=/nuts/discovery/
       - NUTS_DISCOVERY_SERVER_IDS=dev:HomeMonitoring2024
       - NUTS_DIDMETHODS=web
+    networks:
+      - orca-shared-services
   nutsnode-init:
     image: badouralix/curl-jq:alpine
     volumes:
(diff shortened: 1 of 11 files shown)


orchestrator/careplancontributor/importer/importer.go
```diff
@@ -0,0 +1,178 @@
...
+
+	fhirclient "github.com/SanteonNL/go-fhir-client"
+	"github.com/SanteonNL/orca/orchestrator/careplanservice"
+	"github.com/SanteonNL/orca/orchestrator/lib/coolfhir"
+	"github.com/SanteonNL/orca/orchestrator/lib/must"
+	"github.com/SanteonNL/orca/orchestrator/lib/to"
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+func Import(ctx context.Context, cpsFHIRClient fhirclient.Client,
```
dwelman, line 15: I'm wondering if there's a nice way we can clean up these parameters, there are a lot and I worry it will be error prone. Maybe an importer struct?

orchestrator/careplancontributor/service.go
```diff
@@ -949,6 +982,116 @@ func (s Service) tenantBasePath(ctx context.Context) (string, error) {
 	return basePath + "/" + tenant.ID, nil
 }
 
+func (s Service) handleImport(httpRequest *http.Request) (*fhir.Bundle, error) {
```
dwelman, line 985: `HandlerWithTracing` adds basic boilerplate tracing for the request, but if you want more specific details about the handleImport method you'll need to add tracing. I don't know how often this will be used and I know the service has to be launched with a specific flag so maybe not needed. Figured I'd mention in any case

orchestrator/careplanservice/service.go
```diff
@@ -1484,6 +1503,58 @@ func (s *Service) createFHIRClient(ctx context.Context) (fhirclient.Client, erro
...
+		),
+	)
+	defer span.End()
+
+	tenant, err := tenants.FromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if httpRequest.Header.Get("Content-Type") != "application/fhir+json" {
+		return nil, coolfhir.BadRequest("Content-Type must be 'application/fhir+json'")
```
dwelman, line 1522: If you want these specific errors traced you'll need to do something like:
```
span.RecordError(err)
span.SetStatus(codes.Error, err.Error())
```

dwelman (approved):

PR #797, opened 2025-08-27
Made the msc-viewer load automatically if the home monitoring is in progress

jdksloan:


zidad (approved):

jdksloan:
> Enable this on test/acc/prd as well?

Indeed busy looking into it now

deployments/dev/docker-compose.yaml
```diff
@@ -55,6 +55,7 @@ services:
       NEXT_PUBLIC_BASE_PATH: /frontend
       SUPPORT_CONTACT_LINK: mailto:support@example.com #This link will be shown at the bottom of the error.tsx page if configured
       PATIENT_VIEWER_URL: https://msc-viewer-url/ehr/patient # When set, this will add a link to the task details page to open the patient in the viewer
+      NEXT_PUBLIC_AUTOLAUNCH_EXTERNAL_APP: true
```
lextas, line 58: do we need a environment variable for this? in what case do we _not_ want to auto launch it?

lextas (approved):

deployments/dev/docker-compose.yaml
jdksloan: Good point. Let's do that instead

reinkrul (approved):

PR #804, opened 2025-08-29
[DEV] Refactor otel errors

dwelman:
Added a convenience wrapper for tracing errors

### DIFF
--- a/orchestrator/careplanservice/service.go
+++ b/orchestrator/careplanservice/service.go
@@ -573,7 +573,7 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
 	localIdentity, err := s.getLocalIdentity(ctx)
 	if err != nil {
 		otel.Error(span, err)
-		coolfhir.WriteOperationOutcomeFromError(ctx, otel.Error(span, err), operationName, httpResponse)
+		coolfhir.WriteOperationOutcomeFromError(ctx, err, operationName, httpResponse)
 		return
 	}
 
--- a/orchestrator/lib/otel/error.go
+++ b/orchestrator/lib/otel/error.go
@@ -3,6 +3,7 @@ package otel
 import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
+	"strings"
 )
 
 // Error records an error on the span, sets the span status to error, and returns the original error.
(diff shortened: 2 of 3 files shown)


orchestrator/lib/otel/tracing_middleware.go
```diff
@@ -26,7 +27,7 @@ func HandlerWithTracing(tracer trace.Tracer, operationName string, handler http.
 
 		span.SetAttributes(semconv.HTTPStatusCodeKey.Int(wrapped.statusCode))
 		if wrapped.statusCode >= 400 {
-			span.SetStatus(codes.Error, http.StatusText(wrapped.statusCode))
+			Error(span, errors.New(http.StatusText(wrapped.statusCode)))
```
reinkrul, line 30: I'm not sure `RecordError()` is right here (as done by `Error`), since there's not really an error?

orchestrator/lib/otel/error.go
```diff
@@ -0,0 +1,26 @@
...
+func Error(span trace.Span, err error, message ...string) error {
+	if err == nil {
+		return nil
+	}
+
+	span.RecordError(err)
+
+	var statusDesc string
+	if len(message) > 0 && message[0] != "" {
+		statusDesc = message[0]
```
reinkrul, line 19: if there's more, maybe just string-join them? Since there's nothing to prevent you from providing multiple (and they're just thrown away)

orchestrator/careplanservice/service.go
```diff
@@ -570,23 +560,20 @@ func (s *Service) handleGet(httpRequest *http.Request, httpResponse http.Respons
...
+		otel.Error(span, err)
 		coolfhir.WriteOperationOutcomeFromError(ctx, err, operationName, httpResponse)
 		return
 	}
 	localIdentity, err := s.getLocalIdentity(ctx)
 	if err != nil {
-		span.RecordError(err)
-		span.SetStatus(codes.Error, err.Error())
-		coolfhir.WriteOperationOutcomeFromError(ctx, err, operationName, httpResponse)
+		otel.Error(span, err)
```
reinkrul, line 575: double

orchestrator/lib/otel/error.go
dwelman: Will do

PR #814, opened 2025-09-02
Feature/846/sof show enrollment

reinkrul:


frontend/app/enrollment/task/components/util.ts
```diff
@@ -0,0 +1,20 @@
...
+// - Task.focus.display
+// - If not set, then: ServiceRequest.code.text
+// - If not set, then: a joined string of ServiceRequest.code.coding display or code
+import {ServiceRequest, Task} from "fhir/r4";
+
+export default function orderTitle(task: Task, serviceRequest: ServiceRequest | undefined): string | undefined {
+    if (task.focus && task.focus.display) {
+        return task.focus.display;
+    }
+    if (serviceRequest) {
```
dwelman, line 12: Nitpick: Couldn't you remove one level of nesting by returning early? 
```
if (!serviceRequest) {
    return undefined;
}
```

orchestrator/careplancontributor/applaunch/smartonfhir/service.go
```diff
@@ -240,27 +281,68 @@ func (s *Service) handleCallback(response http.ResponseWriter, request *http.Req
...
+			},
+		},
+		Name: []fhir.HumanName{
+			{
+				Family: to.Ptr("TODO"),
+				Given:  []string{"TODO"},
+			},
+		},
+	}, tenant, nil
+	//fhirClient := createFHIRClient(ctx, must.ParseURL(issuer.issuerLaunchURL), tokens.AccessToken)
```
dwelman, line 326: Can these comments be removed?

orchestrator/careplancontributor/applaunch/smartonfhir/service.go
```diff
@@ -63,6 +67,7 @@ type Service struct {
 	cookieHandler      *zitadelHTTP.CookieHandler
 	jwtSigningKey      *jose.SigningKey
 	jwtSigingKeyJWKSet *jose.JSONWebKeySet
+	cpsClient          fhirclient.Client
```
dwelman, line 70: Maybe I am missing it, but it doesn't seem like this is set or used?

orchestrator/careplancontributor/applaunch/smartonfhir/service.go
```diff
@@ -240,27 +281,68 @@ func (s *Service) handleCallback(response http.ResponseWriter, request *http.Req
...
+	// but only to inspect existing Tasks.
+	// It expects the patient to exist, using the patient ID from the EHR to look it up in the CPS FHIR instance.
+	// This should probably change in the future, since this relies on patient IDs copied from the EHR to CPS (which should rely on its own IDs instead).
+	var patient fhir.Patient
+	if err := cpsFHIRClient.Read(patientID, &patient); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to read patient resource: %w", err)
 	}
-	return &patient, &practitioner, nil
+	// TODO: Make this the right Practitioner
+	practitionerID := uuid.NewString()
```
dwelman, line 310: We were previously fetching the practitioner and the patient, why are we not using the right practitioner now?

dwelman (changes requested):

orchestrator/careplancontributor/applaunch/smartonfhir/service.go
reinkrul: We discussed with St. Antonius that we can't access the FHIR API, so they'll be providing practitioner info in the launch request (URL) instead. But that doesn't work yet, so for now it's something hardcoded.

PR #975, opened 2026-04-14
Add condition column to task overview table

rdeveen:
In the task overview the condition of the individual tasks are not listed.

image

# Context

# What

A column displaying the condition of the task

# Whom

Medical professionals that use the Epic EHR

# Why

When a patient is enrolled for multiple conditions the medical professional cannot discern the task they need to click when they want to see homemonitoring patient data

# User story

As a medical professional in the hospital
I want to see what conditions the patient is enrolled for
so that I can choose the correct task when I want to see their homemonitoring data


JTeeuwissen (dismissed):
test

PR #1002, opened 2026-08-13
DEV-2847: add authenticated verification app launch

zidad:
## Summary

New `verification` app launch: opens the EHR viewer flow for an enrolled patient without a hospital, so the launch → session → OIDC → viewer chain can be verified in production. Config-gated (`applaunch.verification`), off by default. Existing launches (SMART on FHIR, Zorgplatform, demo) are untouched.

Two entrypoints:
- `POST /verification-launch` — for a workforce application, authenticated with an OIDC bearer token (existing `rp` client; Entra for us). Only registered where OIDC is configured.
- `POST /verification-launch/scp` — for another SCP node delegating a launch, authenticated over Shared Care Planning's own trust. Never delegates onwards.

The body only names the enrollment Task. When another node holds it, the launch is delegated there, so the session, frontend and OIDC issuer stay with the care organization that owns the enrollment — as in a launch from its own EHR. Patient, ServiceRequest and Condition are read server-side from that node's CPS; only the Practitioner name is caller-asserted.

The response is a single-use, 60s launch URL; `GET /verification-launch/{nonce}` creates the session cookie and redirects into the normal frontend flow. All resources are session-cached, so no EHR is needed.

## Context

ZBJ tracking: DEV-2847. Complements #1001 (the Condition ends up in the session, so the viewer receives the `condition` claim like a real launch).

Launch creation logs tenant, task and the initiating user for auditing.

### DIFF
--- a/orchestrator/careplancontributor/applaunch/verification/noncestore.go
+++ b/orchestrator/careplancontributor/applaunch/verification/noncestore.go
@@ -1,0 +1,37 @@
+package verification
+
+import (
+	"time"
+
+	"github.com/SanteonNL/orca/orchestrator/careplancontributor/applaunch/session"
+	"github.com/google/uuid"
+	"github.com/jellydator/ttlcache/v3"
+)
+
+// nonceStore holds prepared launch sessions until the browser redeems them. A nonce is single-use and
+// short-lived, so the launch URL behaves like an authorization code: the authenticated POST proves who
+// asked, the unauthenticated GET only converts that proof into a session cookie.
+type nonceStore struct {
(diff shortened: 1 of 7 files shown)


orchestrator/careplancontributor/applaunch/verification/noncestore.go
JTeeuwissen, line 1: is there a way to not have to write this all ourselves?

zidad: let me check

zidad: simplified

PR #1003, opened 2026-08-14
Never write an OperationOutcome without issues

zidad:
Upstream FHIR servers (e.g. Azure FHIR on 410 Gone) can return an OperationOutcome without any issue entries. The Go models marshal the nil slice as "issue": null, which strict FHIR parsers (Firely SDK) reject — and issue is 1..* in the spec anyway. SanitizeOperationOutcome now backfills a default processing issue, and WriteOperationOutcomeFromError sanitizes issueless outcomes on every status code.

Found via Zorgbijjou datahub: reading a hard-deleted Task through the CPC external FHIR proxy returned 410 with issue:null, crashing the Firely-based consumer.


jochemvankessel (approved):

JTeeuwissen (approved):
if it works, it works.

PR #1008, opened 2026-08-25
DEV-2891 resolve Dependabot alerts across Go and npm projects

zidad:
## Summary

- Go (`orchestrator`, `e2e-tests`): all 40 open Go alerts resolved via `go get -u` + `tidy`; `setup-go` 1.24 → 1.25 to match the directive `tidy` raised; regenerated `vcrclient_mock` for the new Nuts VCR method.
- npm: all three apps to `next` 16.3.2 — `hospital_simulator` 16 → 0 vulns, `frontend` 123 → 62, `viewer_simulator` 36 → 17.
- Dropped `fhir` from `viewer_simulator`: unused (those imports resolve to `@types/fhir`), deprecated, and *bundles* a vulnerable lodash no override could reach.
- `next lint` is gone in Next 16, so `lint` scripts call `eslint` directly and the configs use `eslint-config-next/core-web-vitals` instead of `FlatCompat`.
- `dependabot.yml` never covered the two simulators for npm/Docker


.github/workflows/orchestrator.yaml
JTeeuwissen, line 1: I feel like this should be deployed to test first for a manual e2e validation (create patient -> enroll -> see omnidesk ticket)

JTeeuwissen (approved):

.github/workflows/orchestrator.yaml
zidad: yes

jochemvankessel (approved):

